PR #1, opened 2015-04-09
removed `is` from `is must be...` message

fourq:
I didn't remove from the lib/ dir since I assumed that dir is built as part of the build process.


jquense:
thanks!

PR #4, opened 2015-05-09
Added thispath to params when formatting error messages

macprog-guy:
Additionally, path is no longer prefixed with 'this'. To keep things exactly as they were, it would be necessary you would need to change the locale.js file, which I left alone because I wanted to get rid of the prefix.

### DIFF
--- a/lib/util/validation-error.js
+++ b/lib/util/validation-error.js
@@ -44,9 +44,9 @@ ValidationError.formatError = function (message, params) {
 
   var fn = function fn(_ref) {
     var path = _ref.path;
-    var params = babelHelpers.objectWithoutProperties(_ref, ['path']);
+    var params = babelHelpers.objectWithoutProperties(_ref, []);
 
-    params.path = 'this' + (path ? path.charAt(0) === '[' ? path : '.' + path : '');
+    params.thispath = 'this' + (path ? path.charAt(0) === '[' ? path : '.' + path : '');
 
     return message(params);
   };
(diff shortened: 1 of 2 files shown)


jquense:
Hi there, Thanks for the the PR,

There are are few issues here, the first is that the changes you made are to the compiled code in `/lib/` instead of the source in `/src`, so the change will be wiped out once I build again. the larger issue tho is that I don't think i want to break path in this way, there are those that rely on the current behavior and there isn't much win to changing it, especially since you can get the behavior you want by using a function as the message instead of a string

```js
required(params => `${removeThis(params.path) is a required field`)) // or es5 equivalent 
```

You can also change the defaults by mutating the values of the `yup.locale` object in your own project.

That being said, the raw path probably should be exposed in the message params, I'd probably be fine with a change that keeps `path` intact but adds `rawPath` for the unprefixed path.

macprog-guy:
I've noticed that I changed the compiled code only after I compiled your code for another PR i'm working on that will add a .locale(object) to mixed so that the locale can be passed dynamically. 

I will change this PR so that it adds a rawpath as you have suggested with good reason and of course, I will do this in the source code.

Note that the reason I don't want to use a function or even so many custom messages everywhere is that it tends to make the model more difficult to read. I would prefer to reserve this for those times where the generic localized message is inappropriate.

macprog-guy:
Actually, since it's just one line of code. Could I you add rawpath to the parameters and I will simply drop this PR.

jquense:
Sure, open up an issue just so I don't forget?

PR #17, opened 2015-12-14
Fixed validation for strings

### LINKED ISSUE #16: Incorrect email validation for not required field
I think email validation works incorrect.
Example:
```
yup.string().email().validate('') - invalid! 
```
I think problem in matches check - test: ```value => value == null || regex.test(value)```
In `transform` null values and empty strings transforms to '' and '' does not equals null!

ruscoder:
Fixed #16


jquense:
see my comments in #16 but this approach isn't the right way to address your concern, empty string values should be treated as strings, and not non-existent values. Certainly min/max validations should work with them

PR #18, opened 2015-12-14
String objects should be considered valid strings.

alpacaaa:
Yo, I think this is valuable. The check is not too insane so it might not target IE 6.12beta or some weird version of Safari, but I think it's good enough :D


jquense:
thanks for this, but in order to be consistent bool, and number should also probably have a similar check, Do you think you could add those as well?

alpacaaa:
There you go!

jquense:
thanks!

PR #19, opened 2015-12-16
Pass the path when rejecting a field based on type

simon-engledew:
Hey there!

I'm using yup to validate form inputs and then display errors if they are incorrect.

I was having a little bit of trouble validating ```number()``` ```typeError()```s and grabbing the path of the incorrect input.

After a bit of digging around, it seems that if you add the path to the reject method it becomes available when dealing with validation errors, which is super useful!

Cheers,
~ Simon

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -108,7 +108,7 @@ SchemaType.prototype = {
     let path = state.path
 
     let errors = [];
-    let reject = () => Promise.reject(new ValidationError(errors, value));
+    let reject = () => Promise.reject(new ValidationError(errors, value, path));
 
     if ( !state.isCast && !isStrict )
       value = schema._cast(value, options)
--- a/lib/mixed.js
+++ b/lib/mixed.js
@@ -113,7 +113,7 @@ SchemaType.prototype = {
 
     var errors = [];
     var reject = function reject() {
-      return Promise.reject(new ValidationError(errors, value));
+      return Promise.reject(new ValidationError(errors, value, path));
     };
 
     if (!state.isCast && !isStrict) value = schema._cast(value, options);


jquense:
Hey there. that's for the PR, I actually think that what you need is already their. in an aggregate validation exception the `inner` property on the error contains each individual error, of which each have a path and type property

simon-engledew:
Yup, typically that works! The problem happens with this schema:

```
schema = (
  yup.object({
    size: yup.number().typeError('enter a number').default(0).required('Enter a size')
  })
);
```

If you try and validate:

```
{
  size: 'a'
}
```

The inner error (```ValidationError("enter a number")```) will not have a path.


Here is the complete example:

```
var schema = yup.object({
  size: yup.number().typeError('enter a number').default(0).required('Enter a size')
});

schema.validate({size: 'a'}, {abortEarly: false}).catch(function (errors) {
  console.log(errors);

  _.each(error.inner, function (inner) {
    console.log('path:', inner.path); // -> undefined
  });
});
```

With this PR the inner.path will be correctly set to 'size'

jquense:
Hey there sorry about the delay, I do see your issue here, but Unfortunately your fix only sort of addresses it. The problem is that the `errors` array is still only creating one error, when it should create 3 (with a path)

simon-engledew:
Sorry, but why would it create three errors? There is a value (required) but it is the wrong type (typeError).

src/mixed.js
```diff
@@ -108,7 +108,7 @@ SchemaType.prototype = {
     let path = state.path
 
     let errors = [];
-    let reject = () => Promise.reject(new ValidationError(errors, value));
+    let reject = () => Promise.reject(new ValidationError(errors, value, path));
```
jquense: `errors` here may be more than just the type error, it could also contain the whitelist and black list errors. This works fine when It's just a type error since that is the first that can be thrown (hence only one), but will break when its used in either the blacklist or whitelist code path.

The main "bug" here is that ValidationError creates an aggregate Error when passed an array of ValidationErrors, but not when it is passed an error of error strings

PR #23, opened 2016-01-26
Fixed minor typo in the ojbect.noUnknown error message

simon-engledew:
Just bumped into this! Not a world changing fix but it proves people are using your software. :D


jquense:
thanks!

Issue #27, opened 2016-02-19
yup.mixed().oneOf([...]) is very slow with large arrays


nfcampos:
Hi,

A call to `yup.mixed().oneOf(arr)` takes tens of seconds to return if `arr` has tens of thousands of items.
I think the slowness is due to the add method of your BadSet class taking only one new element at a time, thus making a call to oneOf() require iterating over the BadSet lots of times.
Would you be open to a PR that changed this? I can also add some tests for the BadSet while I'm at it class to make sure I don't break anything.

Thanks for your work on yup, it's great!

jquense:
I am always open to PRs that make stuff faster and add tests. :)

jquense:
also for arrays that size it's probably worth it to check for and use native Set if it exists

nfcampos:
:)
yep i'll look into using native Set. 
Although, if the performance is similar (after i make this change to the BadSet) i think it'd be better to just always use the BadSet, otherwise it's always possible the BadSet and the Set differ in some obscure way that causes a bug in yup only some environments, no?

jquense:
> BadSet and the Set differ in some obscure way that causes a bug in yup only some environments, no?

that's always a concern but we don;t use anything fancy or obscure about the Set api so Its probably fine. In anycase though I think you'll find creation and lookup for the native Set are going to be much faster, making the risk worth it. I'm open to either/both approaches though

nfcampos:
I looked a but more into this and there is one significant difference between native Set and BadSet, how you deal with dates:
http://esnextb.in/?gist=481ad667c6d618bee4f2 (you have to open the console to see the output)
To make native Set behave like that towards dates requires giving up a big part of its speed advantage in creation and lookup.
I'm guessing you'd like to keep that value equality for dates, right?

nfcampos:
An object-based set keeps the value equality and is waay faster than the current array-based :) 
http://jsperf.com/badset/9
(and since I'm using JSON.stringify() to stringify values they keep their types, JSON.stringify(2) !== JSON.stringify('2'), obvious downside is the set can only hold serialisable values, but since that's true of all the types yup supports I'm guessing there's no problem with that?)

jquense:
Yeah that sounds reasonable to me :)

nfcampos:
one last question, you ok with using a getter for the length of the set or is yup compatible with browsers below IE 9?

nfcampos:
[...]

PR #28, opened 2016-02-21
faster implementation of the BadSet class, tests for the BadSet class

nfcampos:
see issue #27

changes:
- added tests for the BadSet class
- reimplemented the BadSet class using an object as the underlying data structure
- changed one existing test to account for different sort order of the new implementation

notes:
- see http://jsperf.com/badset/9 for performance comparison
- the iteration order of an object is undefined in the js spec, so not all browsers do the same. thus, we can no longer rely on BadSet.prototype.values() returning the values sorted by insertion order. in my opinion this is a small price to pay for the huge speed advantage of using an object instead of an array. however, if need be i can modify this implementation to store insertion order and sort the return value of BadSet.prototype.values() before returning it. however, this comes with a performance penalty, and I don't really see any reason why the order should matter for the purposes of validation library

### DIFF
--- a/src/util/set.js
+++ b/src/util/set.js
@@ -1,43 +1,30 @@
-var toString = Object.prototype.toString
-var isDate = obj => toString.call(obj) === '[object Date]'
-
-module.exports =  class BadSet {
+module.exports = class BadSet {
 
   constructor(){
-    this._array = []
-    this.length = 0
+    this._map = {}
   }
 
   values(){
-    return this._array
+    return Object.keys(this._map).map(v => this._map[v])
   }
 
-  add(item) {
-    if(!this.has(item)) 
-      this._array.push(item)
+  get length(){
+    return Object.keys(this._map).length
+  }
 
-    this.length = this._array.length
+  add(item){
+    this._map[stringify(item)] = item
   }
 
   delete(item){
-    var idx = indexOf(this._array, item)
-    if( idx !== -1) 
-      this._array.splice(idx, 1)
-
-    this.length = this._array.length
+    delete this._map[stringify(item)]
   }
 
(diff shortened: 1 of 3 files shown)


src/util/set.js
```diff
@@ -1,43 +1,30 @@
-var toString = Object.prototype.toString
-var isDate = obj => toString.call(obj) === '[object Date]'
-
-module.exports =  class BadSet {
+module.exports = class BadSet {
 
   constructor(){
-    this._array = []
-    this.length = 0
+    this._map = {}
```
jquense: you'd get some amount of improvement from using `Object.create(null)` here as well, since engines optimize those objects as hash lookups

src/util/set.js
```diff
@@ -1,43 +1,30 @@
...
-      this._array.splice(idx, 1)
-
-    this.length = this._array.length
+    delete this._map[stringify(item)]
   }
 
-  has(val){
-    return indexOf(this._array, val) !== -1
+  has(item){
+    return this._map.hasOwnProperty(stringify(item))
```
jquense: if you do use `object.create` this would break you need to do something like `{}.hasOwnProperty.call(this._map, item)`

nfcampos: i guess in that case i could just use the `in` operator, no (it shouldn't have anything other than ownProperties)?

nfcampos:
@jquense i've changed it to Object.create(null) per your suggestion

nfcampos:
@jquense anything missing to merge this in?

jquense:
nope sorry just got lost in the shuffle :) Thanks a lot for the help!

nfcampos:
Thanks :)

Issue #13, opened 2015-10-20
Add .label() option 


alexrogers:
I've found the label method in Joi quite useful https://github.com/hapijs/joi/blob/master/API.md#anylabelname but I've not been able to find an equivalent in yup.  Would be great if you could add it!

jquense:
hey there, I am not against it, but I haven't ever seen a big need for `label()` myself, what is the use case?

alexrogers:
In my case I'm using keys which are named in camelcase like `startDate`.  For the human-readable errors it would nice to display them as 'Start Date'.

jquense:
ahhhh gotcha, Yeah sounds fine, I am a bit short on time at the moment but i'd take PR for it if you want to give it a shot

kjanoudi:
+1 on the request for label, do you have any requirements for a PR to be approved?

jquense:
just that it doesn't add any runtime overhead, and doesn't break stuff :P

PR #31, opened 2016-03-11
Added label() function.

paulgraves:
As requested in issue #13 - 'Add .label() option'

Have added one simple test as well, let me know if you need more.


paulgraves:
I probably shouldn't have committed the built lib files here to be honest...

PR #37, opened 2016-03-25
allow oneOf required validation

itajaja:
I don't know what else removing that check does, I guess there was a reason for it to be there :D wdyt


jquense:
thanks!

PR #39, opened 2016-04-09
add match function

lostrouter:
add a field matching function.


jquense:
hey! thanks for the contribution. I think that ultimately this should be expressable in terms of a [`ref()`](https://github.com/jquense/yup#refstring-path-object-options) which was just added recently to yup (like the joi counterpart). 

I've already make methods like `max()` and `min()` accept refs, but not oneOf/notOneOf which are harder since they use a weird Set object underneath. The options are to either make those accept refs, or jsut add a `match` (needs a new name to not conflict with string.match) like you did here that accepts a ref. I think overall altering oneOf is probably best, but we need to be careful not to regress on: #28

lostrouter:
i'm having a hard time wrapping my head around ```ref()```.  i can see where mostly how to implement it for ```oneOf```, except i'm not sure how to get the context or parent within ```oneOf```

PR #51, opened 2016-06-02
Allow synchronous usage for a schema consisting of only synchronous v…

jacobp100:
…alidators

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -8,15 +8,39 @@ var Promise = require('universal-promise')
   , cloneDeep = require('./util/clone')
   , createValidation = require('./util/createValidation')
   , BadSet = require('./util/set')
-  , Ref = require('./util/reference');
+  , Ref = require('./util/reference')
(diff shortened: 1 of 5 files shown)


jquense:
hey this is is awesome! thanks a lot for taking the time to do this. There are a few issues as is with this approach. The big one is that this breaks when validating objects or arrays synchronously. That is due to the other schema overriding `_validate` in which they assume a promise will be returned. 

In general all the checking and code branching for sync/async really complicates the code and makes it harder to extend which is a feature of the library i'd like to maintain. I've thought about this a bit before and i was thinking that perhaps the best way to handle this is to have (dare I say it) little wrapper around Promise that has a sync resolution path. That way the code internally can use the same API (then, catch) and the values can be unwrapped at the edges of the API in `validate` and `isValid` (or perhaps validateSync, isValidSync)

Does that make sense? thoughts?

jacobp100:
I like the idea of a synchronous promise! There's even a few NPM packages that do this.

PR #52, opened 2016-06-02
Implement sync promises

jacobp100:


jacobp100:
I think this is actually more complicated—the sync still has to be permeated, as `Promise.all` won't keep the species. I did notice that `cast` is synchronous, and I can actually cover my use-case with just that. Is there a reason why casting is sync, but validation is async, and what is the actual use case for async validation?

PR #63, opened 2016-08-30
[added] number methods less, more, notEqual

A11oW:
Add few new methods for number. This useful use for validate ranges:

```
var yup = require('yup')

var schema = yup.object({
  start: yup.number(),
  end: yup.number().more(yup.ref('start'), 'End must be greater than the start')
})

schema.validate({
  start: 3,
  end: 5,
})
```

Tnx @maullerz for help

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -9,86 +9,123 @@ let isNaN = value => value != +value
 let isInteger = val => isAbsent(val) || val === (val | 0)
 
 export default function NumberSchema() {
-  if ( !(this instanceof NumberSchema))
-    return new NumberSchema()
+	if ( !(this instanceof NumberSchema))
+		return new NumberSchema()
 
-  MixedSchema.call(this, { type: 'number' })
+	MixedSchema.call(this, { type: 'number' })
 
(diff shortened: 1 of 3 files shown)


src/number.js
```diff
@@ -9,86 +9,123 @@ let isNaN = value => value != +value
 let isInteger = val => isAbsent(val) || val === (val | 0)
 
 export default function NumberSchema() {
-  if ( !(this instanceof NumberSchema))
-    return new NumberSchema()
+	if ( !(this instanceof NumberSchema))
```
jquense: can you fix the tab spacing here? It should be two spaces, no tabs.

A11oW:
I fixed the tab spacing.

src/number.js
A11oW: I fixed the tab spacing

jquense:
thanks! sorry for the long delay! :|

jquense:
fyi i'm tweaking the names a bit, and I think actually removing `notEqual` since `notOneOf` should already cover that case

PR #65, opened 2016-10-12
Update readme to reflect abortEarly default value

dingbat:
`mixed#validate`'s `abortEarly` option defaults to `true`, not `false`: https://github.com/jquense/yup/blob/master/src/mixed.js#L41. This took me some time to realize because the readme documents otherwise. This PR corrects that mistake!

Thanks for the great library!


jquense:
thanks!

PR #67, opened 2016-10-19
Added ability to validate string with exact length.

jcsmesquita:
Instead of having to set min and max to specify a precise length for a string this pull request adds a length method to string.


jquense:
thanks!

Issue #81, opened 2017-03-13
Fails on casting `null` values with `date().default(...)` schema


code4aman:
Can't cast date field to its default if value is null.
However, [docs statement](https://github.com/jquense/yup#mixeddefaultvalue-any-schema) claims that non-nullable schema should be casted to its default if actual value is undefined or null. Undefined and null with nullable schema behaviors are just as expected.
Here is the at-the-moment quote from master:
> Sets a default value to use when the value is undefined (or null when the schema is not nullable)

Here is [problem sample](https://runkit.com/58c6d864ce9386001415a40d/58c6d8656a2cc90013eb5131) forked from default demo.
Is this the desired behavior?

upd: add quote in case of master changes

jquense:
values are only to default when undefined, not null, which is considered a separate value generally, not a missing value

code4aman:
oh, so docs statement is not clear then?
> Sets a default value to use when the value is undefined (or null when the schema is not nullable)

jquense:
hmm yeah the docs are wrong here. care to PR a fix :) ?

code4aman:
tbh, I liked previous falsy logic, since we have C# as back-end APIs with null as default value for non-nullables

PR #82, opened 2017-03-13
update docs on mixed().default(...) and nulls

code4aman:
#81 fix


code4aman:
tbh, I liked previous falsy logic, since we have C# as back-end APIs with null as default value for non-nullables
...but that is another ~story~ issue

PR #84, opened 2017-03-16
Fix string interpolation 0 issue

toddbluhm:
Fixed an issue where when using yup in the following way 
```js
yup.number().min(0, 'This field must be greater than or equal to ${min}')
```

The string replace method would interpret `0` as `false` in the string replace methods `or` statement and therefore replace with an empty string.

Added some basic test cases for `ValidationError.js` `formatError` method including one checking specifically for this `0` case.

Also fixed up a few more linter issues.

### DIFF
--- a/src/ValidationError.js
+++ b/src/ValidationError.js
@@ -2,7 +2,13 @@
 let strReg = /\$\{\s*(\w+)\s*\}/g;
 
 let replace = str =>
-  params => str.replace(strReg, (_, key) => params[key] || '')
+  params => str.replace(strReg, (_, key) => {
+    const val = params[key]
+    if (typeof val === 'number') {
+      return val
+    }
+    return val || ''
+  })
 
 
 export default function ValidationError(errors, value, field, type) {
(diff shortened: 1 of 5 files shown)


src/ValidationError.js
```diff
@@ -2,7 +2,13 @@
 let strReg = /\$\{\s*(\w+)\s*\}/g;
 
 let replace = str =>
-  params => str.replace(strReg, (_, key) => params[key] || '')
+  params => str.replace(strReg, (_, key) => {
+    const val = params[key]
+    if (typeof val === 'number') {
+      return val
+    }
+    return val || ''
```
jquense: how about just `String(val)`? that would print `"null"` and `"undefined"` but that seems like the right behavior anyway here?

toddbluhm: Not sure. Was just trying to keep the same intent that the original coder had (defaulting to `'`), I just added and exception `0` issue. I can certainly go that route if you would like?

toddbluhm: Though one concern doing `String(val)` instead would be breaking backwards compatibility. This would be a big visibility change...especially if users are passing in undefined values unintentionally and still receiving decent looking results. Hopefully that makes sense.

jbinto:
Just wanted to thank @toddbluhm for the PR. My first ever `yup` schema was literally `price: yup.number().min(0)` and I was wondering why the message was missing the 0.

PR #88, opened 2017-03-30
docs: remove outdated parameters passed to test() function

jbinto:
Just a small update to the docs. I noticed in the [changelog](https://github.com/jquense/yup/blob/master/CHANGELOG.md#090):

> test functions are no longer passed path and context as arguments, Instead they are now values on this inside the test function.

This PR removes an outdated reference to these parameters in the README.


jquense:
thanks!

PR #89, opened 2017-03-31
Fix Promise error when passing null as the second value to .then

toddbluhm:
- Passing a `null` 2nd param to `.then` is not allowed (when using Bluebird
promises, it warns you)
- Return a function instead that re-throws the error to execute the endEarly condition

### DIFF
--- a/src/util/runValidations.js
+++ b/src/util/runValidations.js
@@ -28,7 +28,7 @@ function scopeToValue(promises, value, sync) {
  */
 export function propagateErrors(endEarly, errors) {
   return endEarly
-    ? null
+    ? err => { throw err; }
     : err => {
         errors.push(err);
         return err.value;


jquense:
long time coming, but why isn't `null` allowed?

toddbluhm:
So I guess technically its javascript so you can pass whatever you want to it, but some promise libs (like bluebird) give errors/warnings when you pass in a non-function because they are assuming that if you pass something in, you probably mean for it to be a function. Its also better practice to propagate errors by re-throwing them, rather then swallowing them. 

For whatever reason I had issues getting errors with this piece of code until I made this change. It could simply be the promise lib I was using, but `bluebird` is pretty popular so it would be nice if it worked with that lib. Having said that, I no longer work at the company that uses `yup`, so I am some what indifferent as to whether this gets merged or not. I would like to see it merged, but its your call 😄

morgs32:
This appears to have been resolved. You can close @toddbluhm @jquense ?

toddbluhm:
So I took a look at this again to see if I could address the issues and figure out a way to either 1) prevent the warning from happening or 2) fix it so that we don't have stack trace issues. 

1) I still couldn't figure out a good way of preventing the warning (both `null` and `undefined` cause it), and globally turning off the warning is just bad practice. 

2) I then took a stab at the stack trace issue when I noticed that I don't have any stack trace issues. My error stack traces are identical (at least as far as I could see, maybe I missed something) whether I use an `(err) => { throw err; }` or `null`. So re-throwing the error does not modify the original stack trace. That makes sense to me because the stack trace is attached to the `err` object itself and the `throw` keyword does not modify that value.

PR #94, opened 2017-05-23
Sync

jquense:


ctavan:
@jquense what are your plans with this?

I'd be really interested in having an option to perform sync validation as well as async validation: My usecase would be to be able to do async validations on the server but running the sync subset of these validations on the browser as well in conjunction with [redux-form](http://redux-form.com/7.0.3/examples/fieldLevelValidation/) which only supports sync field-level validations.

Is there anything I could help you with to get this merged?

jquense:
@ctavan yes there is! I'm happy to get this merged but it adds a lot of complexity and i'd let to get some feedback from folks using it. I don't have a lot of use cases right now for sync validations so it's a bit hard to test. IF you want to give it a shot in a project or hit it hard with examples and see if it falls over that would really help a lot towards feeling confident that this isn't going to be a mistake to merge :P

ctavan:
Cool. I'll rebase the branch and test it in my projects.

retorquere:
I would be *very* interested in this. I'm looking for a validation library I can use in conjunction with lokijs and async validation would be very complicated if possible at all.

retorquere:
I'm currently using ajv which is great except for validation of native dates.

ctavan:
I will continue working on this as soon as #115 has been merged.

PR #99, opened 2017-06-28
dont fail when getOwnPropertyDescriptor returns undefined

neruchev:
We get the error `Can not read property 'get' of undefined`, which occurs when `getOwnPropertyDescriptor` returns `undefined`.

A similar error was [fixed](https://github.com/hapijs/hoek/commit/14113f7bc17a5850588624e8b8157df60ebc6617) in the module `hoek`.


jquense:
thanks!

PR #105, opened 2017-07-18
Support custom locale

grsabreu:
In order to work now, the user should create a file like this and initialize it before all yup usage.

```
import { setLocale } from 'yup/lib/locale'

setLocale({
 email: { invalid: 'Não é válido' },
})

// then after...
import yup from 'yup'
```

I'm going to add tests and update README after you guys approve this!
Thanks for the great package btw. Let me know if any changes are necessary before merging.


src/locale.js
```diff
@@ -1,13 +1,19 @@
+import { merge } from 'lodash/fp'
```
jquense: I think you can use `Object.assign` instead of a lodash module

jquense:
Thanks for taking the time to do this 👍

src/locale.js
grsabreu: I went for lodash.merge because it'll merge deeply, so the user can change just one entry and let the rest default for the current ones.

jquense: you aren't doing any deep merging though, since  it's merging each namespace individually. I'd prefer to maintain that as well since `merge` is not the tiniest part of lodash

grsabreu: Yeah, that's what I wanted, to merge each namespace inside the locale object.
image

But I think you are right, I'm going to just make an or case instead of merging. (`export let string = customLocale.string || {...}`).

grsabreu: Ah, now I got what you meant haha
I'm going to make the adjustments soon

grsabreu: thanks

grsabreu:
Done @jquense

guilhermedecampo:
In need of this.

README.md
```diff
@@ -138,6 +139,19 @@ schema.cast({
 // => { name: 'jimmy', age: 24, createdOn: Date }
 ```
 
+### Using a custom locale dictionary
+```js
```
jquense: can you flesh this out a bit more, explanation of the code, why you need to do this, etc.

yurijean:
:+1:

README.md
grsabreu: yes, I'll do it soon

grsabreu: done @jquense

README.md
```diff
@@ -138,6 +139,25 @@ schema.cast({
...
+
+setLocale({
+  string: {
+    email: 'Invalid',
+  },
+})
+
+// Now use Yup schemas AFTER you defined your custom dicionary
+```
+Why do you have to set the locales before you use Yup at all?
```
jquense: I was think more about what this does generally, not why it needs be first. something like: 

"Allows you to customize the default messages used by Yup, when no message is provided with a validation test."

grsabreu: oh ok

jquense:
thanks!

PR #115, opened 2017-08-17
Cleanup build

ctavan:
As a preparation for my work on the `sync` branch I found a few things which made it hard for me to collaborate on this repo, mainly the fact that the `lib` directory is checked into git.


test/yup.js
```diff
@@ -7,7 +7,7 @@ import { object, array, string, lazy, number } from '../src';
 describe('Yup', function(){
 
   it('should export', function(){
-    require('../lib') // eslint-disable-line global-require
+    require('../src') // eslint-disable-line global-require
```
jquense: this doesn't test the same thing. This is supposed to be a smoke test for the final compiled module not being broken.

jquense:
`./lib` is checked in to make it much easier to depend on the repo from github. you can jsut link the package in your package.json without any additional build steps and it works. Otherwise you have to build and link yourself to depend on it that way. I'm curious what about it being checked in makes it hard? You don't have to worry about merges or anything, just check in whatever the output is

ctavan:
I have a few concerns with built code being checked in:

* One of my main concerns is that the diffs in github pull requests always contain a lot of unrelated, autogenerated code which makes them harder to review and focus on the essential parts.
* Then having `/lib` not gitignored messes up search & replace for me (can of course fix that locally).
* I tend to use `git rebase` to get my pullrequests up-to-date instead of `git pull`, and rebasing autogenerated stuff onto a history of autogenerated changes very often leads to endless (pointless) conflicts.

As for including the package via git: I think typically it shouldn't be necessary to include the master-branch `master` should usually more or less correspond to the latest published package. The major use-case I see for referencing the module from `github` are development branches. But there again you can simply check in the `lib` directory for your development branch and that doesn't need to end up in master…

Maybe this is too much motivated by my own personal workflows, I can just tell for myself that it would be much easier for me to contribute to this library, if `./lib` wasn't checked into git…

jquense:
>  But there again you can simply check in the lib directory for your development branch and that doesn't need to end up in master…

sure, that just strikes me as more complicated :P you are gonna run into all the same issues but also have to make sure it doesn't get merged in 

I'm fine with this, I don't usually need to reference stuff from github anymore so it works for me to remove it. I just wanted to make sure you weren't inadvertently making it harder for yourself to try out like the `sync` bits in a project.

if it works for you, its fine for me

ctavan:
Yeah, I would prefer to remove the lib directory.

If you really want to test a particular version, since npm now offers free scoped public packages, you can still register your own "organization" at npm an publish packages as prerelease-versions there for free: Plus you get a scenario that is even closer to the real package (because you also run the actual `release` script). Example: https://www.npmjs.com/package/@contentpass/yup

So yeah, I'd prefer to get rid of the `lib` directory…

jquense:
@ctavan sorry for the delay can you just clean up the branch and i'll merge this in

test/yup.js
ctavan: Should we remove this test then?

ctavan:
@jquense I just rebased my branch and I also added node 8 to the travis.ci build file.

I was wondering if I should remove the former smoke-test entirely? I think it doesn't make a lot of sense any more.

ctavan:
@jquense how about merging this one now? I would really love to contribute some substantial improvements to the library (like support for sync validations) but the difficult build-setup is currently holding me back.

Having this one merged would make it a lot easier for me to provide clean and easy-to-review pull requests for the changes that I intend to contribute.

In this PR I have btw fixed some linting issues and added the linting to the the `test` stage (which is executed by travis) as well.

Thanks in advance!

PR #138, opened 2017-11-01
Added oneOfType validation

tkvw:
Hello I created a oneOfType validation, it works like PropTypes.oneOfType. I think it's a useful addition. 

I use it for argument validation of a function. An argument can either be an object matching a shape or a string to lookup this object.

### DIFF
--- a/src/alternatives.js
+++ b/src/alternatives.js
@@ -4,7 +4,9 @@ import { alternatives as locale } from './locale.js';
 import isSchema from './util/isSchema';
 import typeOf from 'type-name';
 import { SynchronousPromise } from 'synchronous-promise';
-
+import createValidation from './util/createValidation';
+import printValue from './util/printValue';
+import runValidations from './util/runValidations';
 
 let promise = sync => sync ? SynchronousPromise: Promise;
 const invalidMarker = {};
@@ -19,14 +21,83 @@ export default function AlternativesSchema() {
 
 inherits(AlternativesSchema, MixedSchema, {
 
-    _typeCheck(value) {
-        return true;
+    _validate(_value, options = {}) {
+        let value  = _value;
+        let originalValue = options.originalValue != null ?
+            options.originalValue : _value
+
+        let isStrict = this._option('strict', options)
+        let endEarly = this._option('abortEarly', options)
+
+        let sync = options.sync
+        let path = options.path
+        let label = this._label
+
+        if (!isStrict) {
+            value = this._cast(value, { assert: false, ...options })
+        }
+        // value is cast, we can check if it meets type requirements
+        let validationParams = { value, path, schema: this, options, label, originalValue, sync }
+        let initialTests = []
+
+        if (this._oneOfType)
+            initialTests.push(this._oneOfType(validationParams));
+
+        return runValidations({ validations: initialTests, endEarly, value, path, sync })
+        .then(value => runValidations({
+            path,
+            sync,
+            value,
+            endEarly,
+            validations: this.tests.map(fn => fn(validationParams)),
+        }))
+    },
+
+    resolveSchemaUsingValue(value){
+        if(this._oneOfTypeOptions){
+            return this._oneOfTypeOptions.find(typeOption =>{
+                return typeOption.isValidSync(value)
+            })||this
+        }
+        return this;
+    },
+
+    isType(type){
+        if(this._oneOfTypeOptions){
+            return this._oneOfTypeOptions.find(it => it === type);
+        }
+        return false;
+    },
+
+    cast(value, options = {}) {
+        let resolvedSchema = this.resolveSchemaUsingValue(value);
+        let result = resolvedSchema._cast(value, options);
+
+        if (
+            value !== undefined &&
+            options.assert !== false &&
+            resolvedSchema.isType(result) !== true
+        ) {
+            let formattedValue = printValue(value);
+            let formattedResult = printValue(result);
+            throw new TypeError(
+                `The value of ${options.path || 'field'} could not be cast to a value ` +
+                `that satisfies the schema type: "${resolvedSchema._type}". \n\n` +
+                `attempted value: ${formattedValue} \n` +
+                ((formattedResult !== formattedValue)
+                    ? `result of cast: ${formattedResult}` : '')
+            );
+        }
+
+        return result;
     },
 
     oneOfType(enums, {
         message = locale.oneOfType,
         cast,
     }={} ) {
+        var next = this.clone();
+
         var validTypes = enums.map(schema => {
             if (schema !== false && !isSchema(schema))
                 throw new TypeError(
@@ -36,14 +107,15 @@ inherits(AlternativesSchema, MixedSchema, {
             cast = cast || schema;
             return schema;
         });
-        return this.test({
+        next._oneOfTypeOptions = validTypes;
+        next._oneOfType = createValidation({
             message,
             name: 'oneOfType',
             test(value) {
-                const {options:{sync},path} = this;
+                const {options:{sync}, path} = this;
                 return validTypes.reduce((valid, schema)=>{
                     if(!schema.validate) return valid;
-                    return schema.validate(value, {sync,path})
+                    return schema.validate(value, {sync, path})
                                  .then(validated=> validated || valid)
                                  .catch(()=> valid);
                 }, promise(sync).resolve(invalidMarker))
@@ -59,5 +131,6 @@ inherits(AlternativesSchema, MixedSchema, {
                 });
             }
         });
+        return next;
     }
 })
--- a/src/alternatives.js
+++ b/src/alternatives.js
@@ -4,7 +4,7 @@ import MixedSchema from './mixed';
 import { alternatives as locale } from './locale.js';
 import isSchema from './util/isSchema';
 import typeOf from 'type-name';
-import { SynchronousPromise } from 'synchronous-promise';
+import { SynchronousPromise } from './util/SynchronousPromise';
 
 let promise = sync => sync ? SynchronousPromise : Promise;
 const invalidMarker = {};
@@ -18,6 +18,9 @@ export default function AlternativesSchema() {
 }
 const defaultShouldUseForCasting = () => true;
 
+let thenable = p => p && typeof p.then === 'function' && typeof p.catch === 'function'
+const isPromiseOrType = (v, schema) => schema.isType(v) || thenable(v)// || isWrapper(v);
+
 inherits(AlternativesSchema, MixedSchema, {
     _typeCheck(v) {
         if (this._oneOfTypeOptions) {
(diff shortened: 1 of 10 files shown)


jquense:
Hey, this is really neat, and something i've wanted to add for a while. I'm not super sold on the idea of explicitly needing to pass a `cast` schema or using the first one, I think ideally you want to use the "best" one as determined by trying to validate against each schema and using the first one that matches. The problem there tho is that you can't validate during a `cast` because cast is sync and validate is async. Perhaps it makes sense to try casting each in strict mode and using the first that doesn't error?

src/mixed.js
```diff
@@ -139,7 +139,12 @@ SchemaType.prototype = {
   },
 
   cast(value, options = {}) {
-    let resolvedSchema = this.resolve(options)
+    let resolvedSchema = this.resolve(options);
+
+    if(this._oneOfTypeSchema){
```
jquense: I think this will be neater and easier to implement as a new schema type instead of extending `mixed`, Joi uses `alternatives` maybe we can copy them there? 

That way this logic can be implementing on top of the base logic and maintain the underlying functions like the conditional schema resolution

src/util/createValidation.js
```diff
@@ -46,6 +50,43 @@ function createErrorFactory({ value, label, resolve, originalValue, ...opts}) {
     , { params })
   }
 }
+export function createOneOfTypeValidation(options) {
```
jquense: I don't think you need to create a new helper here. test functions have access to the schema instance (via `this`) so if you store the possible types on the schema you can access them in a test and call out to the inner types, without needing to add a new helper here

tkvw: This is what I had first, but sync validation fails with : 
``` 
Error: Validation test of type: "oneOfType" returned a Promise during a synchronous validate. This test will finish after the validate call has returned
```

```
    next._oneOfTypeError = createValidation({
        message,
        test(value) {
          const {sync} = this.options;
//            console.log("sync? ",this);
            return validTypes.reduce((valid, schema)=>{
                if(!schema.validate) return valid;
                return schema.validate(value, {sync})
                             .then(validated=> validated || valid)
                             .catch(()=> valid);
            }, promise(sync).resolve(invalidMarker)).then(result=>{
                if(result === invalidMarker){
                    return this.createError({
                        params:{
                            values: validTypes.map(subType => typeOf(subType)).join(', ')
                        }
                    });
                }
                return result;
            });
        },
```
Any ideas?

jquense: yeah i think the check that throws that error doesn't check if the promise is a sync one. Maybe we should flag the promise object as synchronous so the error can check and ignore in this case

src/mixed.js
tkvw: 👍

tkvw:
I don't have a deep understanding of the API, but I think I've changed the code to what you've suggested, I you have any other suggestions, let me know. 
Currently it uses `isValidSync` (I should probably catch Error here) to lookup the best fit schema to cast the value. I don't run it in strict mode, because the user probably wants a value like '10' to resolve to 10. So the order in which the types are applied matters.

tkvw:
BTW: those failing Travis tests are known? Can I ignore them?

src/alternatives.js
```diff
@@ -0,0 +1,136 @@
...
+export default function AlternativesSchema() {
+    if (!(this instanceof AlternativesSchema))
+        return new AlternativesSchema()
+
+    MixedSchema.call(this, { type: 'alternatives' })
+}
+
+inherits(AlternativesSchema, MixedSchema, {
+
+    _validate(_value, options = {}) {
```
jquense: most of this logic is copied from mixed, which you don't need to do. Check out how Array cast/validate is implemented because it's similar to what you want to do here, e.g. proxy the calls to a "sub schema"

src/alternatives.js
```diff
@@ -0,0 +1,136 @@
...
+    resolveSchemaUsingValue(value){
+        if(this._oneOfTypeOptions){
+            return this._oneOfTypeOptions.find(typeOption =>{
+                return typeOption.isValidSync(value)
+            })||this
+        }
+        return this;
+    },
+
+    isType(type){
```
jquense: this is should probably be left unimplemented

src/alternatives.js
```diff
@@ -0,0 +1,136 @@
...
+        if(this._oneOfTypeOptions){
+            return this._oneOfTypeOptions.find(it => it === type);
+        }
+        return false;
+    },
+
+    cast(value, options = {}) {
```
jquense: you want to override `_cast` not `cast`

src/alternatives.js
```diff
@@ -0,0 +1,136 @@
...
+            path,
+            sync,
+            value,
+            endEarly,
+            validations: this.tests.map(fn => fn(validationParams)),
+        }))
+    },
+
+    resolveSchemaUsingValue(value){
+        if(this._oneOfTypeOptions){
```
jquense: this does work but will break if there are any async validations which isn't ideal...Maybe it's enough to just use cast like you use validate for the validate case e.g.

```
for(let schema of this._oneOfTypeOptions) {
  try { 
     return schema.cast(value) 
  } catch (err) { 
     continue //failed to cast
  }
}
```

src/alternatives.js
tkvw: Shouldn't I do some type of checking if the type is valid? I see you use
```
    _typeCheck(v){
        return Array.isArray(v)
    },
```
in array, I gather this is to bail out as soon as possible right? I could check if any of the supplied types are applicable here.

src/alternatives.js
tkvw: I like this solution, will change

tkvw: What should happen when a cast is attempted on a value which is not supported? Throw error?

jquense: yeah that's what this does: https://github.com/jquense/yup/blob/master/src/mixed.js#L152

tkvw:
I think I beginning to understand the logic. Please see last changes. I gave the developer an option to include a function to decide if a schema should be used when casting, and each cast is checked if it is a valid type. I can't get it to work without TypeCheck, because the default mixed casting behaviour is to  return the value if the type is allowed on schema. I just implemented in `_typeCheck` now.

src/alternatives.js
```diff
@@ -0,0 +1,144 @@
...
+            const {
+                schemas
+            } = this._oneOfTypeOptions;
+            return !!schemas.find(it => it.isType(v));
+        }
+        return false
+    },
+
+    _validate(_value, options = {}) {
```
jquense: This is still reimplementing to much logic, in fact you probably don't need to implement this at all now that `oneOfType` is implemented in terms of a normal validation. just letting the base mixed.prototype._validate get called will work.

src/alternatives.js
```diff
@@ -0,0 +1,144 @@
...
+        if (this._oneOfTypeOptions) {
+            const {
+                schemas,
+                shouldUseForCasting = defaultShouldUseForCasting
+            } = this._oneOfTypeOptions;
+            for (let schema of schemas) {
+                // It's difficult to decide what type we want to use for casting
+                // the value. Ultimately, it's up to the developer so allow the developer
+                // to decide using `shouldUseForCasting` (defaults to for all types).
+                if(shouldUseForCasting(schema, value, options)){
```
jquense: I'm not sure I understand this toggle, ideally the schema you use cast should be the same one used to validate, you don't want to cast a value to a Date and then validate it as a number for instance

src/alternatives.js
tkvw: So I create the validation in a temp variable in next, how can I make sure the validation runs without implementing _validate ? Should I assign it directly to a func in next?

        var next = this.clone();
        next._oneOfType = createValidation({....

jquense: you don't need to use createValidation at all, `next.test()` should be fine

src/alternatives.js
tkvw: Ideally yes, but there is loss of information. See this example:
```
      inst.validateSync('10',{strict:true}).should.equal('10');
      inst.cast('10').should.equal(10)
```
The first is handled by `string()` and the latter by `number()`. So I created an escape hatch for developers who want to influence this. 
Better would be to validate before cast and throw an Error if a promise is used to validate and inform that validate should be used for casting with promise validators. Or to have cast return a promise as well, but that would be breaking and a bad idea.

src/alternatives.js
tkvw: Ah, that cleans up 👍

tkvw:
So I added another alternative: PromiseOrType, which was a real head breaker because it didn't work. Finally I discovered that SynchronousPromise resolves a real Promise and that is bad for Yup. I created an issue on their repo https://github.com/fluffynuts/synchronous-promise/issues/8, but for now I simply copied in SynchronousPromise and resolved this issue.

jquense:
Yeah the sync promise thing I was aware of, but we throw an error at the only point a user can pass a real promise so it isn't really an issue...Ideally i'd prefer not to vendor in the code even if it's not perfect.

I'm not sure I understand `promiseOrType` method, but I tend to think it adds to much complexity and inconsistency with the API. It's easy enough for a user to resolve promise values before passing them to yup so I don't think there is a lot of value in adding that as an explicit option, especially if it mucks with the general cast/validate flow. I'm happy to chat a bit more about it, but lets keep this PR focused on adding the one method and making sure it's rock solid before we do feature addons

tkvw:
I'll refactor this feature in another PR. I am developing a sync/async validation for arguments and shape of stamps. Arguments and shape should comply with a yup schema (like proptypes) and because arguments can be validating objects themself (promises), this feature makes a lot of sense. Just let the schema take care of it.

Because of this I just discovered another issue.  The docs state that the `validate` promise resolves to the validated value (or whatever is returned from test). This is not the case for arrays. I think validate of array should resolve to a product of all validated values of the array. It *looks* like yup is doing this at the moment, but what actually is happening is that each value of the array is cast (in none strict modus), all validations are run and if none failed the casted value of the array is returned, thus not the product of all results of the validated values. I don't know if this is by design?  

I'll already found a solution for this, depending on whether this is by design or not we could either fix it or maybe add another method to allow this behaviour. Something like `validateAndCast`

jquense:
>  I don't know if this is by design?

Sorry i'm a bit unsure what the behavior your advocating for is compared to what it does, could you maybe put a little example demonstrating what you think shouldn't happen?

tkvw:
Yeah, so I've misread the docs a bit. All schema tests that I've seen returned the value they've tested,so I assumed this was the value being returned. This is the behaviour I was expecting: 
```
        const apiValidator = mixed().test('apivalidation', '', function(value){
            // This is an api server call
            return new Promise(r => {
                setTimeout(()=> r(value+value),1000)
            });
        });
        return Promise.all([
            apiValidator.validate(10).should.eventually().equal(20),
        ])
```
Which validates and *transforms* the value to validate, which kinda cool I think (and also what I'm looking for, especially because they can run async). 
I'll refactor what I've done and open separate issues/feature requests. Like you've said; lets focus on the `oneOfType` first.

tkvw:
Superseded by https://github.com/jquense/yup/pull/141

PR #140, opened 2017-11-02
Fix Date examples in README.md

HarleyKwyn:
It looks like the date examples in the README are incorrect or at the very least not following the convention of not mutating the original value in transforms. I believe the issues is that instead of re-assigning value to the moment object, the intention was to allocate a new variable by the name of date with the moment object.

Let me know if this is correct. Or if I'm suppose to be following the pattern described in the read me and I'm missing something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,13 +223,15 @@ reach(schema, 'nested["arr"][1].num')
 Adds a new method to the core schema types. A friendlier convenience method for `schemaType.prototype[name] = method`.
 
 ```js
+  var invalidDate = new Date('');
+  
   yup.addMethod(yup.date, 'format', function(formats, parseStrict) {
 
     return this.transform(function(value, originalValue){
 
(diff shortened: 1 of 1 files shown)


marr:
Looks like this probably needs a rebase. Also there is a typo in the doc here https://github.com/jquense/yup#extending-schema-types In the _Creating new Types_ section. 
`next._validFormats = {}.concat(formats);` should probably be: ` next._validFormats = [].concat(formats);`

README.md
```diff
@@ -1050,7 +1052,7 @@ function parseDateFromFormats(formats, parseStrict) {
     if (this.isType(value))
       return value
 
-    value = Moment(originalValue, formats, parseStrict)
+    var date = Moment(originalValue, formats, parseStrict)
```
navarroaxel: could you please use `let` or `const` for variable declaration?

PR #141, opened 2017-11-03
Add oneOfType

tkvw:
Supersedes https://github.com/jquense/yup/pull/138
Refactored oneOfType to an isolated schema.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,7 +39,8 @@ node_modules
 .LSOverride
 
 # Icon must end with two \r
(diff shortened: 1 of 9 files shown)


tkvw:
So, in the end I decided to go with a bit other direction. I created an `or` schema in `mixed`. This allows me to do : 
```
string().or(number()).or(date())
```
You can pull it from https://github.com/jquense/yup/compare/master...tkvw:or if you want.

PR #142, opened 2017-11-03
Added transform for validation

tkvw:
Allows validate to resolve to the results of the validations. This can be requested by adding the option `transform` to the option (see added test). 

```
        const apiValidator = mixed().test('apivalidation', '', function(value){
            // This is an api server call
            return new Promise(r => {
                setTimeout(()=> r(value+value), 1000)
            });
        });
        return Promise.all([
            apiValidator.validate(10, {transform:true}).should.eventually().equal(20),
            array().of(apiValidator).validate([10, 20, 40], {transform:true}).should.become([20, 40, 80]),
            array().of(apiValidator).validate([10, 20, 40], {transform:false}).should.become([10, 20, 40]),
            array().of(apiValidator).validate([10, 20, 40]).should.become([10, 20, 40])
        ])

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,7 @@ Options = {
   abortEarly: boolean = true;
   stripUnknown: boolean = false;
   recursive: boolean = true;
+  transform: boolean = false;
   context: ?object;
 }
 ```
@@ -361,6 +362,8 @@ than after all validations run.
 - `stripUnknown`: remove unspecified keys from objects.
 - `recursive`: when `false` validations will not descend into nested schema
 (relevant for objects or arrays).
+- `transform`: this will make validate to return the results of the validations 
+  instead of the input values. 
 - `context`: any context needed for validating schema conditions (see: `when()`)
 
 ```js
--- a/src/array.js
+++ b/src/array.js
@@ -100,7 +100,8 @@ inherits(ArraySchema, MixedSchema, {
           value,
           errors,
           endEarly,
-          validations
+          validations,
+          transform:options.transform
         })
       })
   },
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -180,6 +180,7 @@ SchemaType.prototype = {
 
     let isStrict = this._option('strict', options)
     let endEarly = this._option('abortEarly', options)
+    let transform = this._option('transform', options)
 
     let sync = options.sync
     let path = options.path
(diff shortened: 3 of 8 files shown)


jquense:
Thanks for all the contributions :)

So this I think conflate two concepts pretty core to the yup api and flow. Casting transforms values and validating tests values against some criteria. Pushing a transform ability into validation breaks that delineation without adding a lot of value. If you need to change or transform an input it can be done during the cast phase...this is a bit more verbose if you have promise values but not byuch since you can do something like:

```js
Promise.resolve(value).then(v => schema.validate(v))
```

Which (unless you are running in strict mode) will transform and validate the value

tkvw:
Your comment and the fact that I understand the api a bit better made me realize I actually needed another schematype: `transform`. This solves all my needs, because your api is so elegant designed it's almost a no-op. (https://github.com/tkvw/yup/blob/schema-transform/src/transform.js)

The branch is here: 
@See https://github.com/jquense/yup/compare/master...tkvw:schema-transform
Maybe this schema would be an interesting addition for yup. 

The cool stuff is in the api schema, I've included an faker transformer test which initiates testing data if you're not working in production. 
https://github.com/tkvw/yup/blob/a3ed9a936916654b8454acb2cb7286d934947630/test/transform.js#L35

tkvw:
Added another schema type to `transform`, `ifInvalid` which handles a schema if the first schema is invalid. https://github.com/tkvw/yup/commit/cabcfaa4542ddd1bc450f51796b46a3f805822c7
This allows us to do some nice things: 
```
    it('do api call if schema data is missing', ()=>{
        const inst = transform().ifInvalid(
            object({
                name: string().required(),
                email: string().required().email(),
            }),
            transform().api,
            fakerGenerator, {
                fields: {
                    name: faker => faker.name.findName(),
                    email: faker => faker.internet.email(),
                }
            }
        );
        return Promise.all([
            inst.isValid(undefined).should.eventually().equal(true),
            inst.isValid(undefined, {skip:true}).should.eventually().equal(false),
            // Only generate in test and development!
            inst.isValid(undefined, {skip:process.env.NODE_ENV==='production'}).should.eventually().equal(true)
        ])
    })
```
or even mimic oneOfType: 
```
    it('mimic oneOfType with ifInvalid', ()=>{
        const inst = transform().ifInvalid(string(), (schema)=> object({name:schema}));

        return Promise.all([
            inst.isValid('de').should.eventually().equal(true),
            inst.isValid({name:'de'}).should.eventually().equal(true),
            inst.validate({name:'de'}).should.eventually().become({name:'de'}),
        ])
    });
```

tkvw:
For this issue I created a new `modify` schema for mixed, object and array to add this behaviour. I have a usecase for resolving promised values using the schema and to `modify` the input using promises (instead of doing this using cast, sync)  You can find it here: https://github.com/jquense/yup/compare/master...tkvw:modify
Maybe useful for posterity

Issue #50, opened 2016-05-18
number.integer check is not working


morewings:
Please fix typo at https://github.com/jquense/yup/blob/master/src/number.js#L72 in order to make number.integer check work

jquense:
what typo?

morewings:
| instead of ||

среда, 18 мая 2016 г. пользователь Jason Quense написал:

jquense:
That's not a typo, thats a bitwise or operator, its there because if you `| 0` a number you get an integer (because bitwise operators only work on integers).

morewings:
Anyway it's not working for me.

среда, 18 мая 2016 г. пользователь Jason Quense написал:

morewings:
And if I change to || it's working.

morewings:
All validations are working here, except integer. And when I change to || it starts working.
`[`stock_${isin}`]: yup.object({
      transferAmount: yup.number()
        .default(defaultValue)
        .max(totalValue, `You don't have enough ${name} stocks`)
        .positive('Please provide positive number')
        .integer('Please provide integer'),
    })`

jquense:
please provide a repo demonstrating it not working, changing that to `||` is definitely wrong, its not doing what you think it is.

morewings:
Here you go. Strict didn't helped or maybe I'm doing something wrong?
https://gist.github.com/morewings/abdc6163669ed7a5adab62edc7899ef1

jquense:
oops sorry i meant `strict: true`

morewings:
I tried, the result is the same. Can you run this code?

2016-05-18 22:55 GMT+02:00 Jason Quense <notifications@github.com>:

jquense:
nope, can you please reduce reduce it to a yup schema and a value that values when you call validate. also make sure you are using the latest version of yup/react-formal

morewings:
You were right it appeared to be babel error. I'm telling than because it dissapeared after babel package update. Please take my apologies. And thanks for great work.

sebpiq:
I have also issue with `integer` validator : 

```javascript 
v = yup.number().integer()
v.validate(Date.now()).catch((err) => {
  console.error('GOT', err)
}) // -> ValidationError: this must be an integer
```

jozanza:
[...]

PR #147, opened 2017-11-10
Use parseInt instead of | operator for integer validation

jozanza:
Seems like the `|` operator check will return a false negative given a sufficiently large integer. This PR solves #50

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -5,7 +5,7 @@ import isAbsent from './util/isAbsent';
 
 let isNaN = value => value != +value
 
-let isInteger = val => isAbsent(val) || val === (val | 0)
+let isInteger = val => isAbsent(val) || val === parseInt(val, 10)
 
 export default function NumberSchema() {
   if (!(this instanceof NumberSchema))


src/number.js
```diff
@@ -5,7 +5,7 @@ import isAbsent from './util/isAbsent';
 
 let isNaN = value => value != +value
 
-let isInteger = val => isAbsent(val) || val === (val | 0)
+let isInteger = val => isAbsent(val) || val === parseInt(val, 10)
```
jquense: `parseInt` isn't the right thing, here we don't want to coerce the value into a number. The reason the original code doesn't work with large ints, btw, is that JS really only has 32bit ints and the bitwise operators first convert the float to an int then truncate it. This leads to overflows in the large number case. 

I'm a bit torn, the original method is more "correct" even if it does seemingly odd things. If we were gonna change it tho, we should use the `Number.isInteger` logic e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill

jozanza: I think the coercion is fine if strict equality is being used (`===`)

jozanza: But any approach that makes false negatives go away is fine by me 👍

sebpiq: +1 for @jozanza approach, `parseInt` + strict equality FTW

jozanza: I finally realized the flaw of the `parseInt` approach:
```js
9999999999999999999.9 === parseInt(9999999999999999999, 10)
// true
```

I work on adding the `Number.isInteger` logic to this PR when I get a chance @jquense

vytch:
parseInt('123456789a', 10) returns 123456789. Why not using type coercion: '123456789a' *1 . It is also 4 times faster.

PR #171, opened 2018-02-06
Added notRequired method to mixed type

ilgiznice:
Added `notRequired` method so the field can be set as not required.

I have a big form on my project and there are fields that are shown if other fields are equal to something. Applying `.required()` to everyting and then using `.notRequired()` in `.when()` method would be simpler than checking if field depends on other field and if so using different `mixed()` types in `.when()` sections.

For example, I have to use:

```javascript
yup.mixed().when('some-field', {
    is: true,
    then: yup.mixed().required(),  
    otherwise: yup.mixed(),  
});
```

Instead of:

```javascript
const mixedType = yup.mixed().required();

mixedType.when('some-field', {
    is: true,
    then: mixedType,
    otherwise: mixedType.notRequired(),
});
```

Keep in mind that `mixedType` can be a lot more complicated and creating a new type from scratch will cause code duplicating.


jquense:
Thanks!

PR #173, opened 2018-02-08
Fix hasInList when list has no refs

Guillaumez:
When upgrading Yup from 0.23 to 0.24 our validators ended in the following error `unknown property values from undefined` in the `hasInList` function in `mixed.js` file.

This is a quick workaround. It seems to work and does not break the tests.

Regards.

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -47,13 +47,10 @@ const addToList = (value, list) => {
 
 const hasInList = (value, resolve, list) => {
   if (list.has(value)) return true;
+  if (!list.refs) return false;
 
   let item;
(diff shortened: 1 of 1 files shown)


src/mixed.js
```diff
@@ -48,7 +48,12 @@ const addToList = (value, list) => {
 const hasInList = (value, resolve, list) => {
   if (list.has(value)) return true;
 
-  let item; let values = list.refs.values()
+  let item;
```
jquense: maybe simplify this a bit and add `if (!this.refs) return false` above this line so we return early for these cases?

Guillaumez:
Your last upgrade seems to fix the problem :) Thx

PR #187, opened 2018-03-16
Squash bluebird promise warnings

morgs32:
About not passing in a function for error handler. Any reason not to?

This is a follow up on #89 to resolve merge conflict

### DIFF
--- a/src/util/runValidations.js
+++ b/src/util/runValidations.js
@@ -28,7 +28,7 @@ function scopeToValue(promises, value, sync) {
  */
 export function propagateErrors(endEarly, errors) {
   return endEarly
-    ? null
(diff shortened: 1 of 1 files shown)


jquense:
this would change the stack trace of the error which isn't ideal. I'd really like to understand _why_ bluebird is complaining about this, does passing `undefined` explicitly fix the issue?

morgs32:
Would it change the stack trace? I didn't account for that. 
In the meantime I see there's a way to configure bluebird to quit the warnings.
And some explanation of why bluebird does it:  https://github.com/twilio/twilio-video.js/issues/107#issuecomment-303807393

PR #192, opened 2018-03-23
Upgrade toposort dependency

zauni:
This resolves an issue with uglify where the toposort.array method gets removed from the compressed file.


jquense:
thanks!

PR #198, opened 2018-04-17
fix getDefault so it respects the context passed into it

Tadwork:


Tadwork:
getDefault was calling a non-existent function `_resolve`. I renamed it to resolve and added two tests to test the existing functionality

jquense:
thanks!

Issue #165, opened 2018-01-15
Supporting this.parent.parent call


sundriver:
Hi there,
 Could you let me know how can one traverse the schema in the test method and reach a grandparent (this.parent.parent). this.parent works fine but this.parent.parent === undefined.

Thanks,

jquense:
there isn't a way to do this at the moment. only the current parent is stored

PsiXdev:
Is there a way to validate array of objects in an object?
`{min: 10, values: [{moreThanMin: 11}]}`

sergioviniciuss:
I know that this is being worked on the #201 (a WIP), but do we have a workaround for it? I need to display a field depending the value of another field and I couldn't find a way to access the info because I cannot access a sibling node from its parent node :(

jquense:
The only workaround would be to stick your top level value in `context` when you validate and reference that in the test

sergioviniciuss:
thanks @jquense

yanzou:
```js
// my data structure
{
   ipVersion: 4,    
   tunnel: {
       gateway: {
            ip: 'xxxx',        <--- IPv4 or 6 depend on top level  `ipVersion`
            mac: 'xxx'
       }
   }
}


.when('ipVersion', {  
   is: (value,  originalValue) => {   <-= why not pass the full 'values' so we can get the value I want?
        return originalValue.ipVersion === 4;
    },
     then: ipv4(),
     otherwise: ipv6()
 })
```

since I am seeing the `originalValue` this there in the conditional.js already: 
![image](https://user-images.githubusercontent.com/7699445/81996739-a759c280-9602-11ea-9077-c8cfea61628f.png)


And, this use case seems to be a very universal requirement IMO, 
(I have seeing several similar issue like https://github.com/jaredpalmer/formik/pull/522, https://github.com/jaredpalmer/formik/issues/506, https://github.com/jaredpalmer/formik/issues/503),  why can't we have this in Yup so we don't have to make some hacky workaround?

jquense:
@yanzou The only thing keeping anyone from having it is someone writing it. If you want to see it in yup send a PR

yanzou:
@jquense how about those 2  PR 
https://github.com/jquense/yup/pull/201 
https://github.com/jquense/yup/pull/556
I am seeing people trying to fix the issue but no more grogress

FrankyCTY:
> The only workaround would be to stick your top level value in `context` when you validate and reference that in the test

Too good

PR #200, opened 2018-04-18
Add the ability to get paths relative to a schema

Tadwork:


Tadwork:
@jquense I am pretty sure this is going to be a contentious PR. The purpose of this PR is to allow refs to handle _relative_ paths such as '.../key' so that they can find data relative to the field they are resolved in. Merging this would help for #165 , as well as for cases where you want to make a schema conditional on values in an array item when the field in question is nested.

Tadwork:
Closing because I tried to rewrite my email address and it broke everything

PR #201, opened 2018-04-18
Add the ability to get paths relative to a schema no rewrite

Tadwork:
@jquense I am pretty sure this is going to be a contentious PR. The purpose of this PR is to allow refs to handle relative paths such as '.../key' so that they can find data relative to the field they are resolved in. Merging this would help for #165 , as well as for cases where you want to make a schema conditional on values in an array item when the field in question is nested. (reopening a new PR that does not have a mutilated git history... :( )

### DIFF
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -1,24 +1,11 @@
 import { getter } from 'property-expr';
-import makePath from './util/makePath';
+import { isRelativePath, default as makePath } from './util/makePath';
 
 let validateType = d => {
   if (typeof d !== 'string')
     throw new TypeError("ref's must be strings, got: " + d);
 };
-function getParentPath(path) {
-  const objectIndex = path.lastIndexOf('.');
-  return path.substring(0, objectIndex);
-}
-function getRelativePath(key, path) {
-  // remove '../' from the beginning of the key
-  key = key.slice(3);
-  path = getParentPath(path);
-  const isRelativePath = key.indexOf('../') === 0;
-  if (isRelativePath) {
-    return getRelativePath(key, path);
-  }
-  return makePath`${path}.${key}`;
-}
+
 export default class Reference {
   static isRef(value) {
     return !!(value && (value.__isYupRef || value instanceof Reference));
@@ -35,7 +22,7 @@ export default class Reference {
     this.key = key.trim();
     this.prefix = prefix;
     this.isContext = this.key.indexOf(prefix) === 0;
-    this.isRelativePath = this.key.indexOf('../') === 0;
+    this.isRelativePath = isRelativePath(this.key);
     this.path = this.isContext ? this.key.slice(this.prefix.length) : this.key;
     if (!this.isRelativePath) {
       this._get = getter(this.path, true);
@@ -54,11 +41,8 @@ export default class Reference {
     const { context, parent, value } = options;
     let refValue;
     if (this.isRelativePath) {
-      const relativePath = getRelativePath(
-        this.key,
-        getParentPath(options.path),
-      );
-      const _get = getter(relativePath, true);
+      this.path = makePath`${options.path}.${this.key}`;
+      const _get = getter(this.path, true);
       refValue = _get(value);
     } else {
       refValue = this._get(this.isContext ? context : parent || context || {});
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -1,25 +1,23 @@
 import { getter } from 'property-expr';
+import makePath from './util/makePath';
 
 let validateType = d => {
   if (typeof d !== 'string')
     throw new TypeError("ref's must be strings, got: " + d);
 };
 function getParentPath(path) {
-  const arrayIndex = path.lastIndexOf(']');
   const objectIndex = path.lastIndexOf('.');
-  let index = arrayIndex > objectIndex ? arrayIndex : objectIndex;
-  path = path.substring(0, index);
-  return path;
+  return path.substring(0, objectIndex);
 }
 function getRelativePath(key, path) {
   // remove '../' from the beginning of the key
   key = key.slice(3);
-  const isRelativePath = key.indexOf('../') === 0;
   path = getParentPath(path);
+  const isRelativePath = key.indexOf('../') === 0;
   if (isRelativePath) {
     return getRelativePath(key, path);
   }
-  return path.length ? path + '.' + key : key;
+  return makePath`${path}.${key}`;
 }
 export default class Reference {
   static isRef(value) {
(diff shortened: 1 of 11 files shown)


jquense:
hey thanks for the PR. I'm not opposed to this at all but there are a few cases that need thinking about. The main reason for the current behavior is that it's a lot possible to resolve a schema if you have it. Think about this case:

```js
const schema = object({
  relative: boolean(),
  a: object({
    prop: mixed().when('../relative', {
      is: true,
      then: mixed().required(),
    });,
  }),
});

yup.reach(schema, 'a').validate(someobject) // how do I make this work
```

Previously all you needed was the direct parent if anything to resolve the refs now you need some unknown amount of ancestors, which makes resolving and using an arbitrary subschema hard. And i bring this up because it's really relevant to form librarys (like react-formal) that need to reach through to sub schema

Tadwork:
Fair point... that being said the parent is also not always available. For example :
```
const inst = object({
      someKey: boolean(),
      prop: string().when('someKey', {
        is: true,
        then: string().required(),
      })
    });
```

`await reach(inst, 'prop').validate()` wont work because someKey is not available now because only 'prop' is getting validated
you can get the conditionality to work correctly by passing `value` as an option to `validate` i.e. `await reach(inst, 'prop').validate(undefined,{value:{someKey:true}})`

But you are correct in pointing out that this breaks down when using _relative_ paths because once `reach` has finished resolving the child schema it loses track of the parents` path
when `validate` is called. Since `getValue` was computing the path on every call it would fail (_even when the `value` was passed into validate) in the following case:
```
const prop = ;
const inst = object({
	relative: boolean(),
	a: object({
	prop: mixed().when('../relative', {
		is: true,
		then: mixed().required(),
		})
	})
});
// prop is required, but it fails to find `relative` one level up because it thinks it is at the root level (no parent path is passed in)
await reach(inst, 'a.prop')
	.validate(undefined, {value:{relative:true}}).should.be.rejected();
```

I pushed an additional commit that should handle this by storing the `path` and `_get` when the path is passed in (at `reach`) 
and then subsequently using that in `getValue` when no (parent) path is passed in to a relative `Ref`

Does this successfully fix the issue you raised?

Tadwork:
I added another change this morning that fixes an issue I found where relative paths were not getting resolved if the _parent_ was resolved through reach

Tadwork:
@jquense just wanted to follow up on this... if this isn't something you want to merge in its current form or you just don't like the way it was implemented I can always try again with a bit more direction about how you would want a feature like this to work

src/Reference.js
```diff
@@ -15,34 +16,38 @@ export default class Reference {
...
+  cast(value, options) {
+    return this.getValue(options);
   }
 
-  getValue(parent, context) {
-    let isContext = this.isContext;
-    let value = this._get(isContext ? context : parent || context || {});
-    return this.map(value);
+  getValue(options) {
+    const { context, parent, path, value = {} } = options;
```
jquense: It feels like adding another property here is not ideal. I admit I haven't had a chance to really look over everything, but i don't quite understand what `value` is here. It seems like minimally what you need to resolve a negative path is the entire nested object, the current `path` to and then the relative path from there. what about changing `parent` to an array of ancestors? and using that to resolve everything?

src/Reference.js
```diff
@@ -15,34 +16,38 @@ export default class Reference {
...
+    let refValue;
+    if (this.isRelativePath) {
+      this.path = getRelativePath(path, this.key);
```
jquense: I don't love attaching context specific properties to the instance, since schema are supposed to immutable. Can you talk through why you went with this approach?

src/mixed.js
```diff
@@ -172,6 +174,8 @@ SchemaType.prototype = {
   },
 
   cast(value, options = {}) {
+    options.value = options.value || value;
+    options.path = options.path || this._path;
```
jquense: this. shouldn't mutate options...couldn't `this._path` be wrong here, if a schema is reused in multiple different other schemas?

src/mixed.js
```diff
@@ -172,6 +174,8 @@ SchemaType.prototype = {
   },
 
   cast(value, options = {}) {
+    options.value = options.value || value;
```
jquense: `value` is unclear name for me, what `options.value` why store it separately?

Tadwork:
@jquense I just wanted to update you that I am still working on this. thank you for your patience

jontansey:
@Tadwork How are you getting on with this? I have a similar use case to you whereby I need to walk up the parent and after spend a long time digging about on this repo have arrived here. Anything I can do to help you out?

Tadwork:
@jontansey I have not had time to work on this lately, feel free to take a stab at it if you want.

vincentjames501:
I would love to see something like this get merged. I need to validate 

```
const messageSchema = object().shape({
  name: string().mustBeUniqueValidatorHere()
});
const rootSchema = object().shape({
  messages: array().of(messageSchema)
});
```

I want to set an error on each conflicting name in the array and there isn't a clear way to do that as `this.parent` doesn't go to the value in `rootSchema`.

vincentjames501:
BTW, there is a temporary work around but not a perfect solution for all use cases (especially if you only want/need to go up one parent):

```js
// Pass the root data to context
schema.validateSync(data, {context: {data: data}})
// In your test, you can just access the root data
.test('myValidator', 'my validation message', function (value) {
  const data = this.options.context.data;
  // Do what you need to do with the root data
})
```

geocine:
Where did you invoke this?
> ```js
> schema.validateSync(data, {context: {data: data}})
> ```

MaxHMoyano:
What happened with this issue?

wemakeweb:
Hey @Tedwork are still working on this? Do you need some help finishing implementing this? We are very interested interested to get this merged.

bryanmorganoverbey:
For some trying to access relative paths from their validation schema, you can use the new 'from' feature.
https://github.com/jquense/yup/pull/556
yup
.string()
.test('name', 'errorMessage', function(value) {
        const { from } = this;
        // 'from' will contain the ancestor tree.
      })

bryantobing12:
> For some trying to access relative paths from their validation schema, you can use the new 'from' feature. #556 yup .string() .test('name', 'errorMessage', function(value) { const { from } = this; // 'from' will contain the ancestor tree. })

is this type safe ?

zakaoai:
Still waiting for a solution ?

I can't resolve a problem that involve parent with when and array : 

`yup.object()
    .shape({
            myParentValue: yup.string(),
            myArray: yup.array()
                .of(yup
                    .object()
                    .shape({
                        myArrayObjectField: yup.string()
                            .when('myParentValue', {
                                is: 'aValue',
                                then: (schema) => schema.required()
                            })
                    })`

Issue #177, opened 2018-02-21
Any value that starts with a digit is validated as number


dontub:
The number validator validates any number that starts with a digit, e.g. `1a` is accepted. This seems to be a problem of the [value to number conversion](https://github.com/jquense/yup/blob/2502f9801ef8ef539d612b374757b6925262ccdb/src/number.js#L19):
```
let parsed = parseFloat(value);
if (this.isType(parsed)) return parsed;

return NaN;
```
From https://www.ecma-international.org/ecma-262/6.0/index.html#sec-parsefloat-string:
> `parseFloat` may interpret only a leading portion of _string_ as a Number value; it ignores any characters that cannot be interpreted as part of the notation of an decimal literal, and no indication is given that any such characters were ignored.

What about changing the code to this:
```
let parsed = parseFloat(value);
if (parsed == value) return parsed;

return NaN;
```

jquense:
hey there, I'd be concerned that a change like this will have effects where strings that should be coerced aren't such as `' 100 '` but maybe that is an unreasonable worry. I agree that the current behavior is a bit looser than would be expected :/

dontub:
To allow leading/trailing whitespace value could be trimmed before if it's a string:
```
if (typeof value === 'string') value = value.trim();
```

jhilden:
Facing the same issue.
In my concrete use case, user enters `'4,6'` (German decimal syntax), which will incorrectly be parsed to just `4`.  It would definitely be great if there would maybe be a nice way to inject your own casting logic.

My current workaround is:
```
yup.addMethod(yup.number, 'delocalize', function () {
  return this.transform(function (currentValue, originalValue) {
    return parseFloat(originalValue.replace(',', '.'))
  })
})

yup.number().delocalize()
```
So, far it seems to be working for me.

Another option could be to add a regex (e.g. `/^\d+(?:[.,]\d?)$/`) to the library that ensures the user actually entered a really valid number and not only something that can "somehow" be cast to to a number.

Something like `yup.number().ensureNumberFormat()`

What do you think?

jquense:
> It would definitely be great if there would maybe be a nice way to inject your own casting logic.

it seems like you've hit on the intended api for doing that already. I'm not sure what would be nicer? If you want it to be the default I usually do something like `yup.localizedNumber = () => number().delocalize()`

I'm also happy to add a regex based transform if someone want's to PR, but it should maintain the style and structure of the existing api, which perfers methods returning new schema, over setting schema wide defaults

[...]

PR #208, opened 2018-05-08
Fixes issue #177

kongdigital:
Also adds support hex decimals
Fixes issue #177

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -6,7 +6,6 @@ import isAbsent from './util/isAbsent';
 let isNaN = value => value != +value;
 
 let isInteger = val => isAbsent(val) || val === (val | 0);
-// let isInfinite = val => Math.abs(val) === Number.POSITIVE_INFINITY;
 
 export default function NumberSchema() {
   if (!(this instanceof NumberSchema)) return new NumberSchema();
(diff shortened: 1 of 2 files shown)


kongdigital:
@jquense Any feedback?

bluedaniel:
Is this library still being maintained?

jquense:
Thanks for jumping there @bluedaniel with that very helpful comment

bluedaniel:
Sorry @jquense but I thought it was a valid question seeing as though the PR is almost 2 months old.

src/number.js
```diff
@@ -6,6 +6,7 @@ import isAbsent from './util/isAbsent';
 let isNaN = value => value != +value;
 
 let isInteger = val => isAbsent(val) || val === (val | 0);
+// let isInfinite = val => Math.abs(val) === Number.POSITIVE_INFINITY;
```
jquense: can we remove the commented code?

src/number.js
```diff
@@ -17,7 +18,10 @@ export default function NumberSchema() {
       if (this.isType(value)) return value;
 
       let parsed = parseFloat(value);
-      if (this.isType(parsed)) return parsed;
+      if (this.isType(parsed) && isFinite(value)) return parsed;
+
+      if (typeof value === 'string' && value !== '' && this.isType(+value))
+        return +value;
```
jquense: Sorry for the delay, I've been trying to think what the best option here is in terms of fixing number parsing more generally.

Rereading everything, it seems like `Number(value)` is the best bet (as you have here) and we remove the parseFloat case. What do you think?

src/number.js
kongdigital: 👍

src/number.js
kongdigital: ActuallyI think it has a problem now that I've revisited it. I realised there is no test for whitespace characters.

MaxMilton:
Hey i'd love to see this feature, is there anything I can do to help get it merged?

jquense:
thanks!

MaxMilton:
❤️ Thanks for the merge! It was holding us back on a big project.

PR #215, opened 2018-05-16
add babel/runtime dependency

### LINKED ISSUE #214: undeclared production Babel dependency in v0.25.0
The npm package includes a babel dependency that is undeclared in package.json

specifically, line 3 of /lib/index.js:

```js
var _interopRequireDefault = require("@babel/runtime/helpers/builtin/interopRequireDefault");
```

## Steps to reproduce

Just initialize a new npm repo and add `yup` as a dependency - it can't be imported in node (I'm running node v9.11.1 but I don't think that's related)

```
$ yarn init
$ yarn add yup
...
$ node
> require('yup')
Error: Cannot find module '@babel/runtime/helpers/builtin/interopRequireDefault'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:548:15)
    at Function.Module._load (internal/modules/cjs/loader.js:475:25)
    at Module.require (internal/modules/cjs/loader.js:598:17)
    at require (internal/modules/cjs/helpers.js:11:18)
```

This appears to be breaking type checking in Flow because it can't resolve the module, although my dev build does appear to work.

mmcgahan:
Fixes #214 

This should probably be treated as a short term fix - a better solution might be to revise the publish/build configuration to it inserting the `require('@babel/runtime...')` call - might be related to `runtime: false` on [this line](https://github.com/jquense/yup/blob/049e39ffe799654c6325b53c312560262ddd6ad6/rollup.config.js#L11)? You probably have a better sense of what you want to do with your published build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "sinon-chai": "^3.0.0"
   },
   "dependencies": {
+    "@babel/runtime": "^7.0.0-beta.47",
     "fn-name": "~2.0.1",
     "lodash": "^4.17.10",
     "property-expr": "^1.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -650,6 +650,13 @@
     "@babel/plugin-transform-react-jsx-self" "7.0.0-beta.47"
     "@babel/plugin-transform-react-jsx-source" "7.0.0-beta.47"
 
+"@babel/runtime@7.0.0-beta.47":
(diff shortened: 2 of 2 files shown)


jquense:
Thanks!

himdel:
This is problematic for packages using babel 6 - does `@babel/runtime` have to be in dependencies as opposed to devDependencies.

PR #217, opened 2018-05-16
fix string.matches to throw error with correct message if second arg is omitted

musou1500:
In code below, error messag to be a empty object `{}`.

```js
const schema = string().matches(/(hi|bye)/);

// default message is expected in message
schema
  .validate('nope')
  .catch(err => console.log(err));
```


jquense:
thanks!

PR #220, opened 2018-05-18
imports on ES6 change in README

luispeerez:
https://github.com/jquense/yup/issues/214

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,11 @@ You define and create schema objects. Schema objects are immutable, so each call
 
 __try it out using tonicdev! https://tonicdev.com/570c52590a85f71200eb09ba/yup__
 
+NOTE: If you are usign ES6 imports, import yup this way [issue #214](https://github.com/jquense/yup/issues/214): 
+```js
+import * as yup from 'yup'
+```
+
 ```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -119,6 +119,11 @@ You define and create schema objects. Schema objects are immutable, so each call
 
 __try it out using tonicdev! https://tonicdev.com/570c52590a85f71200eb09ba/yup__
 
+NOTE: If you are usign ES6 imports, import yup this way [issue #214](https://github.com/jquense/yup/issues/214):
```
jquense: I don't think we need to link the issue. 

"When using es module syntax, yup exports everything as a named export"
and
```
import * as yup from 'yup' // for everything
// or only what you need
import { string, object } from 'yup'
```

jquense:
thanks!

luispeerez:
No problem

gruckionvit:
Is this still a suitable way of using yup with named imports rather than `* as Yup` in order to reduce bundle size?

```ts
import * as yup from 'yup' // for everything
// or only what you need
import { string, object } from 'yup'
```

I noticed it's been remove from the readme.

PR #230, opened 2018-05-30
Rename locale keys "less" and "more"

nikita-graf:
Validations `lessThan` and `moreThan` use locale keys `less` and `more`. It seems these two are the only ones which use different naming for a method and locale key. It leads to an error with a `d.ts` file if you want to set up a custom locale. 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yup/index.d.ts#L253

I think it makes more sense to change the naming of locale keys rather than changing types definition.


jquense:
thanks

PR #237, opened 2018-06-25
Lazy load locale

Liooo:
### Problem 

Currently we have to import yup AFTER calling `setLocale()` for the change to take effect, which is not very convenient especially when the app can change locale dynamically during runtime.

### Solution

Reference locale [as lazily as possible](https://github.com/Liooo/yup/blob/1d74e294c54cefc4ee71c6757e5488253c80d538/src/util/createValidation.js#L48-L49), and let `setLocale()` take effect regardless of where yup is imported.

This change should be backward compatible.


Liooo:
@jquense could you check or merge if it's looking okay?

Liooo:
@jquense could you check or merge if it's looking okay?

Liooo:
@jquense 

we need this change in our app, I know you're busy but could you give any feedback about this?

jquense:
Sorry for the delay @Liooo. can you talk a bit why you can't load the locale before yup? At the moment locale data is evaluated at schema creation, not on yup import actually, since that's when the default params are evaluated.

I don't necessarily mind this approach but i'm working to try and reduce the bundle size of yup and this change increases it, tho extremely minimally.

src/array.js
```diff
@@ -115,21 +114,21 @@ inherits(ArraySchema, MixedSchema, {
     return next;
   },
 
-  required(message = mixed.required) {
+  required(message = null) {
```
jquense: i don't think the default to `null` is necesary is it?

src/mixed.js
```diff
@@ -341,13 +346,15 @@ SchemaType.prototype = {
    */
   test(...args) {
     let opts = args[0];
+
+    // TODO: search all method calls and use hash as the argument
     if (args.length > 1) {
-      let [name, message, test] = args;
+      let [name, message, test, localePath] = args;
```
jquense: i wouldn't support this option in anything other than the options object call signature, this is mostly an internal option anyway

src/number.js
```diff
@@ -79,16 +82,21 @@ inherits(NumberSchema, MixedSchema, {
     });
   },
 
-  positive(msg = locale.positive) {
-    return this.min(0, msg);
+  positive(msg = null) {
+    return this.min(0, msg, 'number.positive');
```
jquense: this is a concern of mine, this api does not scale or compose well. Ideally we only want one message argument that can somehow be evaluated lazily. Perhaps instead of `localePath` pass an object with a custom `toString` that resolved to the string value you want?

Liooo:
@jquense 
We just talked about this again today and problem was solved. (using with formik, and didn't recognize that we could pass function to `validationSchema` param, which is evaluated lazily). Sorry for bothering many times.

(Now the only advantage from this PR is may be that we can use `setLocale()` even after rendering the component)

src/array.js
Liooo: right, just for clarity and consistency

jquense:
nice! If you do want to still work on this i'd take some version of it. lazily creating schema is also an approach but probably the less optimal, perf-wise for frequent calls or complex schemas. That really depends on your use case tho, so whatever works best for ya'll 👍

PR #241, opened 2018-07-03
Replace property-expr package to property-expr-csp

Enet:
To satisfy Content-Security-Policy requirements the package `property-expr-csp` was used instead `property-expr`. A new one has the same API, but doesn't cache getters and setters.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@babel/runtime": "^7.0.0-beta.47",
     "fn-name": "~2.0.1",
     "lodash": "^4.17.10",
-    "property-expr": "^1.2.0",
+    "property-expr-csp": "^1.4.0",
     "synchronous-promise": "^1.0.18",
     "toposort": "^2.0.2"
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,7 +22,7 @@ const base = {
     'toposort',
     'fn-name',
     'synchronous-promise',
-    'property-expr',
+    'property-expr-csp',
   ],
 };
 
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -1,4 +1,4 @@
-import { getter } from 'property-expr';
+import { getter } from 'property-expr-csp';
 
 let validateName = d => {
   if (typeof d !== 'string')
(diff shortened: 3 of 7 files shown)


jquense:
Hey there thanks, for the PR, there is already previous discussion on this topic and I'd prefer to take one of the discussed approaches. I'm not comfortable swapping out an important dependency of this library with one I don't control or know anything about.

Enet:
Hi, @jquense. Thank you for the quick response.

Actually `property-expr-csp` differs from `property-expr` by only commit https://github.com/Enet/expr/commit/5025230b8a65097f8d9aa09fd02ce65cfc1ae4a4 and I've published the code as a new package only because the main idea of `property-expr` was broken. So it was silly to make pull request with that commit.

I understand your concerns and have some possible ways to solve the problem:
1.Modify `property-expr` (detect CSP and if so, use the code from `property-expr-csp` or like that). But the library size will be increased.
2. Or place CSP detection straight into `yup` package. If CSP was turned on, `yup` could use another `getter` and `setter` functions.

Or maybe you have some ideas how would it better to do that. I'm ready to implement it, because absolutely don't want to maintain own `yup` fork.

jquense:
I actually like the idea of doing detection and falling back on the slower method, if it's reliable to detect (should just be a try catch around a `eval` call i think?). I do think tho that we should keep the non-eval method closer to what lodash does: cache the path splitting, and implment in terms of a while loop. As have said before i'm happy to take a PR in property-expr with a reasonable similarly performant implementation

le0pard:
Hello @jquense 

Thanks for your library `yup`. It really help us to improve our validations.

We also start getting issues with CSP, which is not allow to do `unsafe-eval` and `unsafe-inline`. Main problem with `try/catch`, because it triggers CSP in browser. If you have CSP `report-uri`/`report-to`, this will create new reports for each usage of the `yup`.

I understand performance needs for `property-expr` to use `eval`.

My question is simple: is anything I can do to improve usage for CSP for yup?

My current idea to add ability by some option to `yup` explain, that it need to use slow method, because `eval` disabled for this environment. 

```js
yup.setConfig({noEval: true})
```

But I need to understand what is your thoughts for such solution.

le0pard:
After some checks looks like options in `yup` will not help. Need fix or replace `property-expr`, because it have https://github.com/jquense/expr/blob/master/index.js#L37-L41 inside file, not in some function.

Issue #135, opened 2017-10-18
Not compatible with strict CSP due to `new Function(..)` in property-expr


novemberborn:
Hi, I just wanted to highlight that this library doesn't work with content security policies where `unsafe-eval` is not allowed. This is because the `property-expr` module compiles functions for property access. See https://github.com/jquense/expr/issues/1 by @harriha.

I suppose the function compilation gives a performance boost, but for me it's more important to have as strict a CSP as possible. I'm not sure whether this should be fixed in https://github.com/jquense/expr/ or whether that dependency should be dropped from `yup` itself.

jquense:
Hey there. I'm probably not gonna change this since the usage of new Function here isn't actually unsafe, and eve  if it was its not at all clear how this could possibly be used maliciously given how you'd already have access to execute code way before this point.

jquense:
If we were to change it tho it'd have to be in property-expr. If you want to put something together there with a similar perf profile I'd be happy to take a look

novemberborn:
> I'm probably not gonna change this since the usage of new Function here isn't actually unsafe

The problem is that I don't want to allow `eval` in my app, even if some uses are safe. CSP is all or nothing, which is quite reasonable because it can't tell where the strings being evaluated come from.

jquense:
i understand you, i'm more noting that this isn't a big priority for me and my limited time to work on OSS. I can see that this is a concern and priotity for you and I'd be happy to take a PR that maintains that addresses it if you put one together. thanks!

novemberborn:
Cool, I hear ya 😄 

Unfortunately I'm leaving this project tomorrow, so until I need to use `yup` next I won't have the time to work on this either. But I'm glad we're clear on an approach so that somebody can pick this up.

Thanks for `yup`!

AndyTyurin:
Any solutions?

ctretyak:
It generates too much reports. We need solution or merge this PR

PR #245, opened 2018-07-09
Use lodash/get in place of property-expr/getter due to strict CSP issues

robframpton:
See #135.

I understand that performance is a concern, but lodash's `get` util does do some caching under the hood. Hoping to get this merged as my current CSP is non-negotiable.


jquense:
see: https://github.com/jquense/yup/pull/241  as well as https://github.com/jquense/expr/pull/5

PR #272, opened 2018-08-04
Upgrade to Babel 7 beta 56, using the new runtime paths

### LINKED ISSUE #260: Dependency not found: @babel/runtime/helpers/builtin/objectWithoutPropertiesLoose
I'm seeing the following behaviour with v0.26.0.  

If I revert to v0.25.1, all is well.

```This dependency was not found:

* @babel/runtime/helpers/builtin/objectWithoutPropertiesLoose in ./node_modules/yup/lib/Lazy.js, ./node_modules/yup/lib/util/runValidations.js and 1 other

To install it, you can run: npm install --save @babel/runtime/helpers/builtin/objectWithoutPropertiesLoose
```

renchap:
Babel 7 beta 56 changed the path for some runtime files compared to the previous versions, making the Yup dist files loading a non-existing path.

This changes the requirement to be beta56+, so the correct runtime will be loaded.

@jquense is it possible to release a fix quickly? This breaks for anybody that does not force the `@babel/runtime` version to be `<= beta55`.

On a side note, Babel Beta 56 should be the latest beta, so things should stabilise soon :)

Fixes #260

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
     ]
   },
   "devDependencies": {
-    "@babel/cli": "^7.0.0-beta.47",
-    "@babel/core": "^7.0.0-beta.47",
+    "@babel/cli": "^7.0.0-beta.56",
+    "@babel/core": "^7.0.0-beta.56",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^8.2.6",
     "babel-jest": "^22.4.3",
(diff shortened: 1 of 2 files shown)


thomasobrien99:
@renchap or @jquense would love to use this since the package is currently broken for us. But it looks like this was opened against an already broken version of the package (?), which was fixed by https://github.com/jquense/yup/commit/84849613310128e4f4014505aa0d3ccef6adb647. So if this could be rebased that would be great, merged even better :)

Issue #216, opened 2018-05-16
Unable to resolve "@babel/runtime/helpers/builtin/interopRequireDefault"


outaTiME:
Hi guys, today I update yup to 0.25.0 and im getting the following error:

```
Unable to resolve "@babel/runtime/helpers/builtin/interopRequireDefault" from "node_modules/yup/lib/index.js"
Failed building JavaScript bundle.
```

Rolling back to v0.24.1 fix the issue, any advice?

k7sleeper:
Quick fix: `npm add @babel/runtime`

jquense:
Upgrade please, fix released thanks!

PoriniFoundation:
still getting the error with the quick fix.


W20180805-11:27:49.710(2)? (STDERR) Error: Cannot find module '@babel/runtime/helpers/builtin/interopRequireDefault'
W20180805-11:27:49.710(2)? (STDERR)     at Function.Module._resolveFilename (module.js:547:15)
W20180805-11:27:49.710(2)? (STDERR)     at Function.resolve (internal/module.js:18:19)
W20180805-11:27:49.711(2)? (STDERR)     at Object.require (/home/timo/ticket/simple-todos/.meteor/local/build/programs/server/boot.js:288:32)
W20180805-11:27:49.711(2)? (STDERR)     at makeInstallerOptions.fallback (packages/modules-runtime.js:653:18)
W20180805-11:27:49.711(2)? (STDERR)     at require (packages/modules-runtime.js:244:16)
W20180805-11:27:49.712(2)? (STDERR)     at livedata_connection.js (/home/timo/ticket/simple-todos/.meteor/local/build/programs/server/packages/ddp-client.js:147:30)
W20180805-11:27:49.712(2)? (STDERR)     at fileEvaluate (packages/modules-runtime.js:339:7)
W20180805-11:27:49.712(2)? (STDERR)     at require (packages/modules-runtime.js:238:16)
W20180805-11:27:49.713(2)? (STDERR)     at namespace.js (packages/ddp-client/common/namespace.js:1:300)
W20180805-11:27:49.713(2)? (STDERR)     at fileEvaluate (packages/modules-runtime.js:339:7)
=

hua-lu:
I have the same issue and installing the latest babel/runtime doesn't work for me

hielkehoeve:
[...]

PR #276, opened 2018-08-07
Pin beta packages to specific versions

bbakersmith:
Yup pulls in the latest beta version of the Babel libraries, which sometimes break because they are in Beta. If beta libraries are required, they should be pinned to specific known-good versions. This PR does that.

Addresses the recurring issue here: https://github.com/jquense/yup/issues/216#issuecomment-411033919

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
     ]
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

PR #280, opened 2018-08-09
Update Babel to 7-rc.1

renchap:
Babel 7 is now in the Release Candidate phase 🎉 

This PR updates Babel packages to the new Babel 7-RC.1 version.
Yarn is complaning when installing as `babel-preset-jason` has a constraint on Babel version:
```
warning "babel-preset-jason > @babel/preset-env > @babel/plugin-proposal-object-rest-spread@7.0.0-beta.54" has incorrect peer dependency "@babel/core@>=7.0.0-beta.50 <7.0.0-rc.0".
```

I think your preset should be updated to allow the latest RC.

Babel 7 API should now be stable and the last few beta broke things, so it would be a good idea to upgrade all your packages to the RC version.

### DIFF
--- a/dist/yup.esm.js
+++ b/dist/yup.esm.js
@@ -647,17 +647,17 @@ var trim = function trim(part) {
   return part.substr(0, part.length - 1).substr(1);
 };
 
-function reach(obj, path, value, context) {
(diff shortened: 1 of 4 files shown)


jquense:
thanks for the quick turnaround but i've just finished this myself!

PR #292, opened 2018-08-19
Add function test names to email and url

aytee17:
I noticed that `string.email()` and `string.url()` don't have associated test function names and appear as `undefined` in the `tests` array of `SchemaDescription`. 

This just makes sure that a proper name is sent through the `matches` function.

### DIFF
--- a/src/string.js
+++ b/src/string.js
@@ -89,7 +89,7 @@ inherits(StringSchema, MixedSchema, {
     }
 
     return this.test({
-      name,
+      name: name || 'matches'
       message: message || locale.matches,
       params: { regex },
       test: value =>
--- a/src/string.js
+++ b/src/string.js
@@ -76,14 +76,20 @@ inherits(StringSchema, MixedSchema, {
   matches(regex, options) {
     let excludeEmptyString = false;
     let message;
+    let name;
 
     if (options) {
-      if (options.message || options.hasOwnProperty('excludeEmptyString')) {
-        ({ excludeEmptyString, message } = options);
+      if (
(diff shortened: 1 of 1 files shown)


jquense:
why close this?

aytee17:
Honestly, I thought I did something wrong (this is my first PR changing code) so I closed it out of embarrassment.

dontub:
`matches` must have a name itself if none is provided, probably just "matches". Without name at least the async validation is (correctly) rejected, but without errors. Until this is fixed I have to overwrite the `matches` method.

src/string.js
```diff
@@ -76,14 +76,20 @@ inherits(StringSchema, MixedSchema, {
...
+        options.message ||
+        options.hasOwnProperty('excludeEmptyString') ||
+        options.name
+      ) {
+        ({ excludeEmptyString, message, name } = options);
       } else message = options;
     }
 
     return this.test({
+      name,
```
jquense:
```suggestion
      name: name || 'matches'
```

juice49:
@jquense do you have a schedule for publishing a new version of yup to npm?

I'm relying on this patch in one of my projects, so I've included the yup master branch for now. It would be great to be able to use yup from npm.

Happy to help if there is anything I can do 🙂.

Thank you for this very useful lib!

sawyerh:
Thanks for tackling this @aytee17, I just noticed this too and was about to open a PR. Glad to see this got merged, and eagerly awaiting its release. Our system logs the `name` to help track what type of errors users are running into.

PR #320, opened 2018-09-25
Add missing colon in README.md

Mahfouzer:
I was searching for the url validation when I came across this 
it's no big deal but it's better added


jquense:
Thanks!

PR #344, opened 2018-10-17
Update README.md

bostrom:
Add note about lexical vs bound context in test function for clarity.


jquense:
thanks!

greghaygood:
I just spent a couple hours struggling with why my tests weren't working as I thought they should. Then I figured out it was this very issue, forked the project so I could add a note, only to then find that @bostrom already added it but it wasn't yet pushed to the live site. Cheers to the community!

PR #347, opened 2018-10-17
Changes array.ensure to work on the original value

### LINKED ISSUE #343: array().ensure() not casting single element
The docs for `array.ensure` say that it will cast a non-array into a single element array:

```
array()
  .ensure()
  .cast(1); // -> [1]
```

However in 0.26.6 on Node, the above returns `[]`. It looks like the `transform` call in the array constructor (run with `withMutation`) runs first and sets the value to `null` when it sees it’s not an array. Therefore the `transform` method in `ensure` sees `null` rather than the original value.

I’m happy to make a PR if I could understand what the right solution is for the fix. Are the `transform`s not being called in the right order?

finneganh:
This fixes to match the documentation, which says that a single element
should get cast to an array containing that element. Previously the
transform added in the constructor would convert non-arrays to null
before the ensure transform function would see them.

Fixes #343

WARNING: This is a breaking change. `ensure` now ignores any
transformations added before it, including the transform in the
constructor that parses JSON.

### DIFF
--- a/src/array.js
+++ b/src/array.js
@@ -164,7 +164,10 @@ inherits(ArraySchema, MixedSchema, {
 
   ensure() {
     return this.default(() => []).transform(
-      val => (val === null ? [] : [].concat(val)),
+      // We transform the original value to allow a single element to
+      // get wrapped in an array. The transform added in the Array schema’s
+      // constructor turns non-array, non-JSON-string values into null.
+      (val, orig) => (orig === null ? [] : [].concat(orig)),
     );
   },
 
--- a/README.md
+++ b/README.md
@@ -609,21 +609,20 @@ You can also specify more than one dependent key, in which case each value will 
 
 ```javascript
 var inst = yup.object({
-      isSpecial: yup.bool(),
-      isBig: yup.bool(),
-      count: yup.number()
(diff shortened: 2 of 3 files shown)


finneganh:
It looks like a precommit hook ran Prettier on README.md.

src/array.js
```diff
@@ -164,7 +164,10 @@ inherits(ArraySchema, MixedSchema, {
 
   ensure() {
     return this.default(() => []).transform(
-      val => (val === null ? [] : [].concat(val)),
+      // We transform the original value to allow a single element to
+      // get wrapped in an array. The transform added in the Array schema’s
+      // constructor turns non-array, non-JSON-string values into null.
+      (val, orig) => (orig === null ? [] : [].concat(orig)),
```
jquense: I think it may make sense to use the original value only if `value` is null, so you don't run into issues with values like `'[1,2,4]'`

kdschaffe: @fionawhim I just ran into a need for this fix. I agree with @jquense. I'd think most transforms would want this behavior to keep them chain-able... Can we get this implemented? Thanks!

Something like this:
```
(val, orig) => {
  const value = val || orig

  return (value === null ? [] : [].concat(value))
}

PR #366, opened 2018-11-13
Lazy validator tests

nikita-graf:
Docs don't mention that `lazy` validator is actually called with `context` as a second argument. I've covered this feature with tests and going to make another PR with updated documentation for this validator.


jquense:
Thanks!

PR #369, opened 2018-11-14
Updated lazy function docs

nikita-graf:
A `lazy` function signature is missing a `context` parameter in the documentation. I simply added it to Readme. Not sure if it should be described more detailed.

Related to https://github.com/jquense/yup/pull/366

Please let me know what do you think :)


README.md
```diff
@@ -609,21 +609,20 @@ You can also specify more than one dependent key, in which case each value will
...
+  }),
+});
 
 inst.validate({
   isBig: true,
   isSpecial: true,
-  count: 10
-})
+  count: 10,
+});
```
nikita-graf: This diff is caused by prettier in a pre-commit hook.

README.md
```diff
@@ -267,7 +267,7 @@ inst.cast({ foo: { bar: 'boom' } }, { context: { x: 5 } })
 // { baz: 'boom',  x: 5, { foo: { bar: 'boom' } }, }
 ```
 
-#### `yup.lazy((value: any) => Schema): Lazy`
+#### `yup.lazy((value: any, context: ?object) => Schema): Lazy`
```
jquense: its not actually `context` here but `options` which can contain `context`

PR #392, opened 2018-12-05
feat: ensure should not change the value if it is already an array.

Anber:
It is expected that `array.ensure` keeps original value if the argument is already an array. Otherwise, it causes some problems with shallow comparison.


src/array.js
```diff
@@ -164,7 +166,7 @@ inherits(ArraySchema, MixedSchema, {
 
   ensure() {
     return this.default(() => []).transform(
-      val => (val === null ? [] : [].concat(val)),
+      val => (val === null ? [] : castArray(val)),
```
jquense: lets avoicd the lodash function and go with
```
if (this.isType(val)) return val
return val === null ? [] : [].concat(val)
```

Anber: Oops…
Done.

Issue #359, opened 2018-11-01
Bundle size vs. Install size


shellscape:
Hey there. Stopped in to evaluate the lib after a colleague recommended it. I'm opening this issue to raise a curiosity. While `yup` is much smaller (by about 42kb) than `joi`, it's nearly 3 times larger of an install, and it looks like the dependencies chosen are the culprit.

I noticed that `@babel/runtime` in in `dependencies` rather than `devDependencies`. On the surface that looks like an error, since I couldn't find where the runtime was being used in `dist` nor `lib` on a cursory search.

The big culprit on the install size seems to comes from `lodash`. And while there is a perfectly valid use-case for `lodash` here, it could be said that such a large dependency isn't actually needed. Looks like the primary use for it is: https://github.com/jquense/yup/blob/dd474ff79fb34aee5780dc0aebeac42eb6b88e48/rollup.config.js#L14-L21. And that's cool, but most of those methods are now available as part of ES5/6, and for the ones that aren't, much smaller one-off modules are available which do the same. For the clientside bundles, `lodash` could be used as a polyfill and kept part of `devDeps` instead.

Anyhoo, I'm a fan of lean-and-mean, and this was a curiosity. Cheers 🍺

jquense:
Hey there, I'm usually not much a fan for optimizing install size ahead of runtime size, performance or ease of maintenance. Part of the problem with even trying to optimize this is that choices may actually make things worse overall when not taken together in the context of the ecosystem. 

`@babel/runtime` (seen [here](https://unpkg.com/yup@0.26.6/lib/index.js)) _reduces_ runtime size by providing reusable helpers instead of inline them in each function. For something like `lodash` we do use it as a runtime dependency in a few places and same deal, instead of rewriting a bunch of common utils myself we use lodash, and the main lodash package, because it's very likely that some other package in your app bundle _also_ uses lodash. Overall you get a smaller app bundle because yup uses lodash (in most cases).

jquense:
It should be said that Joi doesn't have to worry about this as much a pure node module, tho likely it's still faster to install using lodash since yarn/npm can dedupe that package and install it only once, again because almost all ndoe/web applications use lodash _somewhere_

jquense:
> If anything, reducing the library's reliance on a massive package and getting up it to speed on the latest developments and not having to be concerned about those utils sounds a lot like it would reduce the maintenance burden, protect against future tech-debt, and future-proof the library.

[...]

PR #402, opened 2018-12-17
feat: add lodash modules as a dependency instead of full lodash

andreoav:


jquense:
Thanks for the PR. The reason we didn't do this originally is that using individual packages is likely to result in _larger_ bundles for consumers since most folks use the plain `lodash` package, so rather than get deduped `has` usage from someone using `lodash` and `lodash.has` you end up with two packages. vs cherry picking from the module like we do here only requires the specific one we need but in the context of the main package so anyone else using `lodash` will not duplicate the functions we're using

andreoav:
@jquense And there is a way to handle ppl that don't use lodash at all?

I use next.js and I have many splits, but everytime I need yup somewhere, I get the full lodash in that split.

jquense:
are you talking about downloading all of lodash or all of lodash being included in in a bundle? The latter won't happen, only the modules we cherry pick are included see #359 for more discussion around install size

divmgl:
To add to this, this is no longer necessary with tree shaking. Instead of bringing in each individual module as a package you can refer directly to a file inside of `lodash` and Webpack should do the rest:

```
import pick from 'lodash/pick'
```

jquense:
We already do that that yeah

PR #403, opened 2018-12-18
fix: cast() does not work well if scheme has stripped field.

Anber:
This is the small fix for #342: cast() returns new value even if nothing has changed, but scheme has a stripped field.


jquense:
thanks!

Anber:
Wow! It was fast! :)

PR #405, opened 2018-12-19
Replace integer check with Number.isInteger

mAAdhaTTah:
This works correctly for large numbers.

Related to https://github.com/jquense/yup/pull/147

### DIFF
--- a/src/number.js
+++ b/src/number.js
@@ -5,8 +5,6 @@ import isAbsent from './util/isAbsent';
 
 let isNaN = value => value != +value;
 
-let isInteger = val => isAbsent(val) || val === (val | 0);
-
 export default function NumberSchema() {
   if (!(this instanceof NumberSchema)) return new NumberSchema();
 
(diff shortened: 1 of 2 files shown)


mAAdhaTTah:
I'm not sure what your policy is for using newer features. I can inline the `Number.isInteger` polyfill from MDN if that's preferred. Or I can update the docs to indicate this is required so the user makes sure it's supported or is polyfilled.

mAAdhaTTah:
Looking at the test failure, I'm not sure why that was considered invalid.

src/number.js
```diff
@@ -94,7 +92,7 @@ inherits(NumberSchema, MixedSchema, {
   },
 
   integer(message = locale.integer) {
-    return this.test({ name: 'integer', message, test: isInteger });
+    return this.test({ name: 'integer', message, test: Number.isInteger });
```
jquense: this still needs to account for missing values and return `true` in those cases

mAAdhaTTah:
Addressed the comment & resolved the conflict.

mAAdhaTTah:
@jquense Bumping this. I think I've addressed your issues. Let me know if you've got other changes.

jquense:
thanks!

pleunv:
Just discovered that `Number.isInteger(Number.MAX_SAFE_INTEGER + 1)` returns `true`, which means that this could cause issues in validation. There's also `Number.isSafeInteger()` which would invalidate anything higher than `Number.MAX_SAFE_INTEGER`. Should Yup be using this one instead?

Issue #409, opened 2018-12-31
Validate object().shape() with ref and abortEarly false


samuelsantia:
Hi there,

I have schema that needs duplicate a field value in other field.
Imagine:

```javascript
const schema = yup.object().shape({
  myProp: yup.string().required(),
  myDupProp: yup.ref('myProp'),
  otherProp: yup.string().required()
})
```

The problem is when I try validate with abortEarly turned off I got the Error `TypeError: promise.then is not a function`

Debugging it I watch that in object _validate method if the field don't exists or don't has validate method it returns true value, breaking the Promise.all of runValidations utils.

I think if field exists it should return Promise.resovle(true) to don't break Promise chain of util.

What do you think?
If you want I can try to do a pull request to fix it (it would be my first one :)

maksimsemenov:
I'm seeing a similar issue when using `.when()` with `abortEarly: false`. I'm getting the same error: `TypeError: promise.then is not a function`.

PR #410, opened 2018-12-31
fix: validate() does not work with ref and abortEarly false. (#409)

samuelsantia:
Commented change on #409 to dont break Promise.all chain


jquense:
thanks!

PR #413, opened 2019-01-03
#411 Add testsParams to SchemaDescription

### LINKED ISSUE #411: Add test `params` to `SchemaDescription`
In some situations, it may be useful to retrieve test `params`. For e.g. given we have some schema creator:
```
const getSchema = count => yup.mixed().test({
  name: 'max',
  exclusive: true,
  params: { count },
  message: 'Must be less than ${count} characters',
  test: value => value == null || value.length <= count,
});

const schema = getSchema(21);
```
With `formik` we then can connect validation with UI but often it's useful to display some help text (in this case we can display current length and max `3 of 21`).
Currently, it's not possible to retrieve `params` passed to yup test.
To improve this we can add an additional field to `SchemaDescription` called `testsParams`.
```
const description = schema. describe();

console.log(description);
// {
//     ...
//     testsParams: { max: { count: 21 } },
// }
```

@jquense what do you think?

desfero:
Closes #411

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -123,7 +123,7 @@ const proto = (SchemaType.prototype = {
     // manually add the new tests to ensure
     // the deduping logic is consistent
     schema.tests.forEach(fn => {
-      next = next.test(fn.TEST);
+      next = next.test(fn.OPTIONS);
     });
 
     next._type = schema._type;
@@ -311,7 +311,7 @@ const proto = (SchemaType.prototype = {
 
   notRequired() {
     var next = this.clone();
-    next.tests = next.tests.filter(test => test.TEST_NAME !== 'required');
+    next.tests = next.tests.filter(test => test.OPTIONS.name !== 'required');
     return next;
   },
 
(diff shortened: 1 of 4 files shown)


src/mixed.js
```diff
@@ -477,15 +477,19 @@ const proto = (SchemaType.prototype = {
...
 
     return {
       type: next._type,
       meta: next._meta,
       label: next._label,
       tests: next.tests
-        .map(fn => fn.TEST_NAME, {})
+        .map(fn => fn.TEST_NAME)
         .filter((n, idx, list) => list.indexOf(n) === idx),
+      testsParams: next.tests.reduce(
```
jquense: i wonder if it's better to make tests a map like this with a value like `{ name, params }`?

desfero: that makes sense and it was my first thought. The reason I go with `testsParams` is to avoid breaking change. If it's ok for you to make it as a breaking change than I can change, it's really small change

desfero: ok, now `params` are with `tests`, can you review? Thanks

jquense:
thanks!

PR #414, opened 2019-01-06
README moreThan and lessThan message interpolation params

minigammi:
Fixed message interpolation params description for `number.moreThan` and `number.lessThan` in README.


jquense:
thanks!

PR #417, opened 2019-01-09
0 is not positive or negative

### LINKED ISSUE #322: 0 is not a positive number
`.positive()` accepts 0, but shouldn't 0 not be considered a positive number?

captDaylight:
fixes #322

### DIFF
--- a/test/number.js
+++ b/test/number.js
@@ -216,8 +216,10 @@ describe('Number types', function() {
 
       v
         .isValid(0)
-        .should.eventually()
-        .equal(true),
+        .should.be.rejected()
+        .then(null, function(err) {
+          err.errors[0].should.contain('this must be a positive number');
+        }),
 
       v
         .validate(-4)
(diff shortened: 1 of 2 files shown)


jquense:
thanks!

test/number.js
```diff
@@ -262,8 +266,10 @@ describe('Number types', function() {
 
       v
         .isValid(0)
-        .should.eventually()
-        .equal(true),
+        .should.be.rejected()
+        .then(null, function(err) {
+          err.errors[0].should.contain('this must be a positive number');
```
jquense: this isn't hit i don't think. `should.be.rejected` will catch the error and return it instead of letting it propagate a believe

captDaylight: yup my bad, I did it without testing locally. Commit coming down the pipeline...

captDaylight:
Tests pass, but I see that "should check integer" test seems to be a duplicate of the positive tests. Should I fix that now?

captDaylight:
good to go!

jquense:
thanks!

captDaylight:
awesome!

PR #421, opened 2019-01-14
unpinning @runtime/babel and publishing pkg.module

activescott:
Resolves an issue that manifests as the following error:
```
Cannot find module '@babel/runtime/helpers/builtin/interopRequireDefault' from 'reach.js'

  at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:221:17)
  at Object.<anonymous> (node_modules/react-formal/node_modules/yup/lib/util/reach.js:3:30)
```
The issue is demonstrated in the following repo https://github.com/activescott/yup-deps-test and I believe is the root of the issue repeatedly reported in jquense/yup#216 .

The specific changes are:

1. Publishing a `pkg.module` in package.json. More about pkg.module is at the following links:
    - https://github.com/nodejs/node-eps/blob/master/002-es-modules.md
    - https://github.com/nodejs/node-eps/blob/4217dca299d89c8c18ac44c878b5fe9581974ef3/002-es6-modules.md#51-determining-if-source-is-an-es-module
    - https://github.com/rollup/rollup/wiki/pkg.module
2. Unpinning @babel/runtime. This seems to have been pinned as part of jquense/yup#276 when @babel/runtime was in beta and I suspect just inadvertantly not unpinned. Since it is not in beta any longer there isn't really a reason to pin. With it pinned to that old version, in all apps that bring along a newer version, yup will drag in it's own old version of @babel/runtime to keep working. This extra runtime isn't always used when babel is bundling a host application that ends up loading this older version of yup. 

Fix 1 allows avoids/works around the issue by allowing babel to transpile the original ES module code rather than the transpiled code that assumes an old version of @babel/runtime.
Fix 2 avoids the issue by just no longer pinning itself to the old version of @babel/runtime. In theory _should_ be safe as long as @babel/runtime follows semver but apparently they don't strictly follow it with these helpers or this issue wouldn't have happened in the first place!


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.26.7",
   "description": "Dead simple Object schema validation",
   "main": "lib/index.js",
+  "module": "src/index.js",
```
jquense: this isn't right, `module` should point to an esm build not source code. The rest of the code here may not be something that can be run in a browser depending on which proposals and bespoke babel plugins i'm running. Practically I do think the code is pretty standard es6 but this isn't safe generally. In general i'd like to handle this via a flat rollup build

activescott: Makes sense. Thank you for the feedback and clear direction. I'll add an esm build w/ rollup and resubmit.

activescott: K. please see my latest updates. I saw you already have a rollup config setup, just not building. I added a `build:esm` script to run rollup, called it from the `build` script and added `/dist/yup.esm.js` to pkg.files. Locally I ran this in a project with `yarn link` and seems to grab the module and work.  Let me know if you have something else in mind or if I'm missing something.

jquense: ya i had started the work to add a rollup config but didn't finish. Honestly i'm not sure if it even works, may need to test it

activescott: @jquense One idea I had was running tests on the es bundle too with a secondary jest config. I reading that react did that and found the article at https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html#running-tests-on-compiled-bundles Let me know if that is a direction you'd like to go to merge this or if you have something else in mind.

PR #428, opened 2019-01-21
Update `mixed.cast` documentation

gopar:
Looks like `mixed.cast` accepts additional parameters that aren't documented.


jquense:
thanks!

PR #433, opened 2019-01-28
Update README w/ proper default for nullable

Nicd:
The [documentation](https://github.com/jquense/yup#mixednullableisnullable-boolean--false-schema) for `mixed.nullable()` shows the typing as:

```
mixed.nullable(isNullable: boolean = false): Schema
```

This would imply that calling `yup.nullable()` sets the schema as **not** nullable. But in fact it does the opposite and sets the schema as nullable. Looking at [the implementation](https://github.com/jquense/yup/blob/master/src/mixed.js#L320), this is true:

```javascript
  nullable(value) {
    var next = this.clone();
    next._nullable = value === false ? false : true;
    return next;
  },
```

If the argument `value` is not provided, it will be `undefined`. That will not match `false`, so `next._nullable` will be set to `true`. So a more proper documented default value is `true`.


jquense:
thanks!

Issue #440, opened 2019-02-01
Allow undefined context, or a set a default context within the schema


kdschaffe:
I found that when a schema uses context,

`.when('$other', (other, schema) => (other === 4 ? schema.max(6) : schema)),`

and a context isn't provided with `schema.validate()`, an error will occur:

`TypeError: Cannot read property 'other' of undefined`

Is there a way to avoid this error, without passing a context into the `validate` call?

A couple ideas: 
1. Set a default context at the schema level 
2. Allow an undefined context (not required to pass a context). Maybe internally, context would default to an empty object, and if context values aren't set, they simply get passed to schema rules as `undefined`. Rules can then handle this as needed.

For reference, this Formik issue might be resolved for some people by making this change in yup: https://github.com/jaredpalmer/formik/issues/503. Although, I think resolving both issues would be beneficial.

jquense:
idk this sounds like good error, albeit not a super clear one. If you define a schema a dependent on context doesn't it make sense to error if you don't provide any context?

kdschaffe:
Thanks for the quick response. In general I think the error is fine. I'm interested in a way to allow the schema to decide wether it should error or not, though. Maybe you want yup to error by default, so option 2 isn't ideal. 

Option 1 would allow the schema to have individual rules handle undefined context, or undefined properties on context, or even better, avoid undefined values altogether. This would help in situations where validation and casting logic is handled by a base class, or wrapper, where most of the time you can use the default context but can overwrite validation/casting and provide a different context when necessary. Hope that makes sense!

Issue #423, opened 2019-01-14
yup.ref(...).transform is not a function


iamvanja:
Trying to create a computed field that depends on another one. Something like: 

```js
...
  originalField: yup.array().default([]),
  computedField: yup.ref('originalField')
    .transform(function (currentValue, originalValue) {
      let value = currentValue || originalValue

      // do some transformation here

      return value
    }),
...
```

Is `.ref` not designed for something like this? If not, is there anything else I could try? 
I would like to keep this logic on a schema level if possible.

Thanks!

jquense:
yes that's right, a `Ref` isn't a schema, its only proxy value. In not sure exactly want you want to do but you can resolve a ref value in a transform:

```
yup.array().transform(function () {
  const otherValue = this.resolve('otherField')
})

```

iamvanja:
That's perfect! Thanks :)

iamvanja:
It looks like `computedField` will run `transform` only when `computedField`'s value is provided. Otherwise it will return what is defined through `default`. Which makes sense, but does not work quite work with my idea of an "imaginary field". 
Thanks for the lightning fast response nonetheless :)

iamvanja:
This can be closed.

vonagam:
For future readers. Right now reference documentation is wrong and second argument for `ref` is actually not an options object, but [`mapFn`](https://github.com/jquense/yup/blob/386eb48928701d5fa39c19b8125d6efac6e6b600/src/Reference.js#L17) which maps value from referenced to resolved one. Example:

```
yup.object({
  x: yup.ref('y', (a) => a + 1),
}).cast({
  y: 3,
}); // => {y: 3, x: 4}
```

jquense:
I HAVE NO IDEA WHEN I ADDED THAT AND DON'T REMEMBER WHY :P

jquense:
Yeah there a few crufty things originally added I think to match joi, but they aren't used and we aren't really trying to match joi anymore. I'm glad for the second pair of eyes on the old code 👍

PR #443, opened 2019-02-03
feat: support self references

vonagam:
`key` cannot be a function since [`validateName`](https://github.com/jquense/yup/blob/d02ff5e59e004b4c5189d1b9fc0055cff45c61df/src/Reference.js#L3) ensures that it is a string.

`isSelf` is not used anywhere and `key="."` isn't supported by `property-expr` getter.

For working self references i opened [pull request](https://github.com/jquense/expr/pull/9) in `property-expr` to support empty string path. (currently `getter("")` works, but `getter("", true)` which is used in Reference.js throws)

### DIFF
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -18,17 +18,21 @@ export default class Reference {
     validateName(key);
     let prefix = options.contextPrefix || '$';
 
-    if (typeof key === 'function') {
-      key = '.';
-    }
-
     this.key = key.trim();
     this.prefix = prefix;
-    this.isContext = this.key.indexOf(prefix) === 0;
-    this.isSelf = this.key === '.';
 
-    this.path = this.isContext ? this.key.slice(this.prefix.length) : this.key;
-    this._get = getter(this.path, true);
+    this.isContext = this.key.indexOf(prefix) === 0;
+    this.isParent = this.key === '';
+    this.isSelf = this.key === '.';
+    this.isSibling = !this.isContext && !this.isParent && !this.isSelf;
+
+    if (!this.isSelf) {
+      this.path = this.isContext
+        ? this.key.slice(this.prefix.length)
+        : this.key;
+      this._get = getter(this.path, true);
+    }
+
     this.map = mapFn || (value => value);
   }
   resolve() {
@@ -36,12 +40,14 @@ export default class Reference {
   }
 
   cast(value, { parent, context }) {
-    return this.getValue(parent, context);
+    return this.getValue(value, parent, context);
   }
 
-  getValue(parent, context) {
-    let isContext = this.isContext;
-    let value = this._get(isContext ? context : parent || context || {});
+  getValue(value, parent, context) {
+    if (!this.isSelf) {
+      value = this._get(this.isContext ? context : parent || context || {});
+    }
+
     return this.map(value);
   }
 }
--- a/src/Reference.js
+++ b/src/Reference.js
@@ -1,54 +1,69 @@
 import { getter } from 'property-expr';
 
-let validateName = d => {
-  if (typeof d !== 'string')
-    throw new TypeError("ref's must be strings, got: " + d);
+const prefixes = {
+  context: '$',
+  value: '.',
 };
 
 export default class Reference {
-  static isRef(value) {
-    return !!(value && (value.__isYupRef || value instanceof Reference));
+  constructor(key, options = {}) {
+    if (typeof key !== 'string')
+      throw new TypeError('ref must be a string, got: ' + key);
+
+    this.key = key.trim();
+
+    if (key === '') throw new TypeError('ref must be a non-empty string');
+
+    this.isContext = this.key[0] === prefixes.context;
+    this.isValue = this.key[0] === prefixes.value;
+    this.isSibling = !this.isContext && !this.isValue;
+
+    let prefix = this.isContext
+      ? prefixes.context
+      : this.isValue
+        ? prefixes.value
+        : '';
+
+    this.path = this.key.slice(prefix.length);
+    this.getter = this.path && getter(this.path, true);
+    this.map = options.map;
+  }
+
+  getValue(options) {
+    let result = this.isContext
+      ? options.context
+      : this.isValue
+        ? options.value
+        : options.parent;
+
+    if (this.getter) result = this.getter(result || {});
+
+    if (this.map) result = this.map(result);
+
+    return result;
+  }
+
+  cast(value, options) {
+    return this.getValue({ ...options, value });
+  }
+
+  resolve() {
+    return this;
+  }
+
+  describe() {
+    return {
+      type: 'ref',
+      key: this.key,
+    };
   }
 
   toString() {
     return `Ref(${this.key})`;
   }
 
-  constructor(key, mapFn, options = {}) {
-    validateName(key);
-    let prefix = options.contextPrefix || '$';
-
-    this.key = key.trim();
-    this.prefix = prefix;
-
-    this.isContext = this.key.indexOf(prefix) === 0;
-    this.isParent = this.key === '';
-    this.isSelf = this.key === '.';
-    this.isSibling = !this.isContext && !this.isParent && !this.isSelf;
-
-    if (!this.isSelf) {
-      this.path = this.isContext
-        ? this.key.slice(this.prefix.length)
-        : this.key;
-      this._get = getter(this.path, true);
-    }
-
-    this.map = mapFn || (value => value);
-  }
-  resolve() {
-    return this;
-  }
-
-  cast(value, { parent, context }) {
-    return this.getValue(value, parent, context);
-  }
-
-  getValue(value, parent, context) {
-    if (!this.isSelf) {
-      value = this._get(this.isContext ? context : parent || context || {});
-    }
-
-    return this.map(value);
+  static isRef(value) {
+    return value && value.__isYupRef;
   }
 }
 
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -380,6 +380,14 @@ const proto = (SchemaType.prototype = {
   },
 
   when(keys, options) {
+    if (arguments.length === 1) {
+      options = keys;
+      keys = '.';
+    } else if (isSchema(keys) || typeof keys === 'function') {
+      options = { ...options, is: keys };
+      keys = '.';
+    }
+
     var next = this.clone(),
       deps = [].concat(keys).map(key => new Ref(key));
 
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -134,11 +134,11 @@ const proto = (SchemaType.prototype = {
     return !this._typeCheck || this._typeCheck(v);
   },
 
-  resolve({ context, parent }) {
+  resolve({ value, parent, context }) {
     if (this._conditions.length) {
       return this._conditions.reduce(
         (schema, match) =>
-          match.resolve(schema, match.getValue(parent, context)),
+          match.resolve(schema, match.getValue(value, parent, context)),
         this,
       );
     }
@@ -147,7 +147,7 @@ const proto = (SchemaType.prototype = {
   },
 
   cast(value, options = {}) {
-    let resolvedSchema = this.resolve(options);
+    let resolvedSchema = this.resolve({ ...options, value });
     let result = resolvedSchema._cast(value, options);
 
     if (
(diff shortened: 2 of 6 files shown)


vonagam:
Found usage for `isSelf` field. Such references should not be added to `_deps` and used in `sortFields`.

vonagam:
Ok, now i understand why `key` was `"."` for self references...

Made `"."` stand for self references again. Updated stuff to work with such references (thanks to presence of lazy type it wasn't hard).

Only public api change is `Reference.getValue` (don't know if this method considered as public though). New first argument for current value. (Can be made as non-breaking change if this is public api by making it third argument, not first)

vonagam:
Three more things:

1) Using self references now it is possible to match joy's signature for `when`:

```js
mixed().when(string(), {then: string().min(5)})
```

Also possible to omit key for self reference:

```js
mixed().when({is: string(), then: string().min(5)})
```

2) In two above examples conditional that was used was not a function or a value but a scheme. This uses `validateSync` of a passed scheme.

3) While writing tests for this stuff i stumbled on problem with `concat`:

```js
yup.number().validateSync([]) // will throw
yup.mixed().concat(yup.number()).validateSync([]) // will return []
```

Even through previous `concat` assigned `_type` to `"number"`, it retained prototype of `mixed`, so no `_typeCheck` from `NumberSchema.prototype` (`ObjectSchema.prototype` has `_cast` and `_validate` redefinitions, so for object it will be even more broken). I fixed it by flipping direction of merging - now target's clone merges into source's clone and source's clone returned.

jquense:
Hey thanks for PR it's a bit tough to follow the work here since it seems like it started off as a small code hygiene fix and ended up adding a few features and changing the API. 

Generally it's be great if we can scope PRs to one change, especially when they change the API so it's easier review and other fixes are blocked behind discussion of unrelated changes. 

For this can we clearly state wat you are trying add and why it's needed? That will help me contextualize the change while review ig as well as understand whether API changes are needed

vonagam:
My end goal - support `mixed.when()` usage like in joi's api (simple case of example specified in my previous comment) https://github.com/hapijs/joi/blob/v14.3.1/API.md#anywhencondition-options

There are 4 different issues:

1) Ability to create a self reference (reference which returns not sibling at parent but value itself). It seems that it was intended at some point or maybe it was possible before since there is `isSelf` property on `Reference`. This is main obstacle in using `when` like in joi. Self references introduce change in signature of `Reference.getValue()` from `(parent, context)` to `(value, parent, context)`. Depending on whenever you consider this method as public api or not it might be a breaking change (it is not documented anywhere in readme, i assume that only reference constructor is part of public api).

2) Shortcut syntax for using `when` with self references. Since before self references were not existent, there was no need for such shortcuts. With them it is more pleasant to work with plus they are present in joi. Shortcuts do introduce non-breaking changes to signature of `mixed.when()`: single argument usage and when first argument is condition.

3) Ability to use scheme as condition `is` in `mixed.when()`. Currently only constant value or function check is supported.

4) Standalone side mission: fix `mixed.concat()`. I presented example and description of the problem in my last comment (last point). Only this issue can be separated into another PR since it does not depend on existence of self references.

Currently there are 3 commits in this PR so it is simpler to look at changes. 
Commit one addresses 1) problem.
Commit two fixes problem 4).
Commit three implements issues 2) and 3).

vonagam:
Additional major refactor of Reference.js file. Changes:

1) Second argument now really is an options object. And it contains `map` function for mapping value from referenced to resolved one. `contextPrefix` option is removed (reason to remove - #423).

2) Allow to look into value deeply, so not only can you use `"."` to get value, but also `".nested.key"` to get a nested value. Just like for context.

3) You can't use key without `$` to get context value. I think it is unexpected behaviour and was not documented. `$` - for context, `.` - for value, empty string - to get parent value, anything else for siblings.

4) Now all possible sources for value are treated equally in a sense that if base for value is absent, then resolved value is `undefined`. Right now context references throw if context is not provided, but do not throw if parent is not provided for siblings. (Partially will close #440)

5) Added `describe()` implementation to `Reference`. Will not throw now if you call `describe()` on a scheme which contains `ref` somewhere.

vonagam:
Out of all source code in this pull request right now self references support take only 10 small lines  (other stuff is tests and refactor with features). What are benefits?

1) Joi migration and parity. No need to migrate to `lazy`, if `when` has same functionality.

1) Easier to read, easier (and less) to write than `lazy`. (`lazy` still may win in really complex scenarios though.)

1) `describe()` does not work with `lazy`. Even if it will in the future, it will never will be able to describe its condition or logic. I think `describe()` might find some useful usage down the line.

1) You can't combine `Lazy` but you can have multiply `when`s. Plus schemas with conditions can combine with others schemas.

1) Performance. In [example](https://github.com/jquense/yup#yuplazyvalue-any--schema-lazy) usage for `lazy` a schema created inside `lazy` body, so for complex schemas it might be faster to just concat to two predefined schemas (like in `when`), than to create one from scratch. (Though i might be wrong on this, this is more of a possibility)

1) `mixed.test()` and `Lazy` are powerful tools and may look like an answer to all, but in reality they move burden (of source code, tests, localisation, possible migration and so on) from library to end user. In some cases it is good, but in others (like `describe()`) not only end users suffers, but library suffers too. `Lazy` is a joker, good to have, but if there is an easier and more canonical way to achieve same result, then this way is better.

Personally, this is (plus lack of `alternatives()`) the issue that made me look into the code of yup. Because i am in process of migration from joi (i want to use same validation library for database models, server params and form). And replacing some schemas which use `when` with `Lazy` feels like regression in usability for me.

src/mixed.js
```diff
@@ -382,11 +381,19 @@ const proto = (SchemaType.prototype = {
   },
 
   when(keys, options) {
+    if (arguments.length === 1) {
+      options = keys;
+      keys = '.';
+    } else if (isSchema(keys) || typeof keys === 'function') {
```
jquense: this is: `when(string(), { then: ... })`? Sorry, I find this to be a confusing api. I realize that Joi has this but it feels like an API wart, i don't want to import. I'm also probably still a 👎 on schema as `is` conditions, but i'll cover that in another reply

test/mixed.js
```diff
@@ -701,6 +712,50 @@ describe('Mixed Types ', () => {
...
+    await inst
+      .validate(4)
+      .should.be.rejectedWith(ValidationError, /must be greater/);
+
+    await inst.validate(5).should.be.fulfilled();
+
+    await inst.validate(-1).should.be.fulfilled();
+  });
+
+  it('should support conditional signle argument as options shortcut', async function() {
```
jquense:
```suggestion
  it('should support conditional single argument as options shortcut', async function() {
```

src/Reference.js
```diff
@@ -1,48 +1,70 @@
...
+    if (typeof key !== 'string')
+      throw new TypeError("ref's must be strings, got: " + key);
 
-  toString() {
-    return `Ref(${this.key})`;
+    this.key = key.trim();
+
+    this.isContext = this.key[0] === prefixes.context;
+    this.isValue = this.key[0] === prefixes.value;
+    this.isParent = this.key === '';
```
jquense: I prefer to leave this out for now. This should be solved more robustly like #201

src/Reference.js
```diff
@@ -1,48 +1,70 @@
...
+    let result = this.isContext
+      ? options.context
+      : this.isValue
+        ? options.value
+        : options.parent;
 
-    if (typeof key === 'function') {
-      key = '.';
-    }
+    if (result === null) result = undefined;
```
jquense: i'd remove this and just check for ` == null` on the getter path below. `null` is semantically different from `undefined` is there a specific reason to do this?

src/Reference.js
```diff
@@ -1,48 +1,70 @@
...
 
-    this.key = key.trim();
-    this.prefix = prefix;
-    this.isContext = this.key.indexOf(prefix) === 0;
-    this.isSelf = this.key === '.';
+    if (this.getter && result !== undefined) result = this.getter(result);
```
jquense: this.getter can't be falsey if the parent check is removed

jquense:
Those are good reasons 👍 I will say tho that API parity with `joi` is really not a goal of `yup`, it's never been meant as a drop-in replacement and even if the api was the same, the internal implementation of casting and validations is so different that you'd likely just get different results than the same schema with joi. I appreciate your perspective coming from joi, and we totally want to make that experience good, even if it's a second order concern. 

I agree with you that declarative affordances like `when` provide a nicer experience, i'm also a heavy user of yup, so well aware of the sharp edges. MY concern is primarily in maintaining clear boundaries between the two apis. yes `lazy` is a low level tool, and at some level of complexity i makes sense to say "use that" instead of "add another overload to when". It's good to chat through these things before moving a head.

In this case i think you're right the code is minimal, and the api is nicer, and easy to support.

src/mixed.js
vonagam: So you prefer `when({ is: check, then: branch })` over `when(check, { then: branch })`? 
What about `when(check, branch)` (will take one line to add)?

src/Reference.js
vonagam: Well, this is about treating all sources equally. If you can get value by `"."`, then path is empty string. If empty string works as path for value, why should it not work for context or parent? Why can't person just get parent or just get context object itself?

This is not about getting values from above a parent like `"../above"`.

Plus what is the status of mentioned PR, is this worked on in any way?

src/mixed.js
jquense: i say go with just `when(options)` and `when(keys, options)` its not a lot of code but a far bit of api complexity for minimal gain

jquense: which is to say i very intentionally trimmed down these overloads when i added `when` originally

src/Reference.js
vonagam: `getter` is falsey if `path` is empty (key = prefix + path). So for key `"."` `path` is empty and so there is no getter.

src/Reference.js
jquense: > then path is empty string. 

That's only b/c you made `.` a prefix :P it wasn't meant as that, but as a single special value. I realize this covers, `'.foo'` as well but i'm not actually sure that is a good idea, it's fine if it falls out naturally tho i guess.

`''` wasn't ever an intentionally supported option, the idea is you can depend on siblings or context.

src/Reference.js
vonagam: Before it was `getter(context || parent || {})` so you could not get `null`. But yeah, now if you have empty path you should be able to get `null`. Fixed.

src/Reference.js
jquense: see comment above ☝️

src/mixed.js
vonagam: Ok, removed those.

src/Reference.js
jquense: > Why can i depend on sibling, but not on parent?

there isn't a reason generally, i think it's a good idea. I would like to scope that as a new feature, and ideally cover more cases instead of half implementing being able to reference parents.

> Why can't i get context as a whole (it might not even be an object)?

it has to be an object, i don't see any value in supporting other types.

vonagam: Mentioned PR is soon will be year old and as i understand its opener is not working on it right now.

`Lazy` does have access to `value`, `context` and `parent`.

Is there any problem with supporting getting a parent value with `""`  right now and then later introducing new stuff?

It might be a breaking change later, but really small and easy one to migrate.

If full feature support will be soon, then migration will be painless, if it will be in a distant future than having something working in meantime is worth it.

Am i missing some angle? Do you see any possible problems with migrations?

jquense: I again, it's a distinct feature and needs to be evaluated on it's own. I don't really feel like adding new additions to the api that offer minimal value and would require a breaking change later as part of an unrelated PR. this isn't even a feature in Joi so i don't think it's a big deal to not support here at this moment.

vonagam: Ok. Removed ability to get a parent.

vonagam: But you still get context object by `"$"`. Is this ok or should be removed too?

src/Reference.js
vonagam: Now it is:

```js
if (this.getter) result = this.getter(result || {});
```

Is there any problem with that?

src/Reference.js
jquense: awesome thanks, I think that's fine for now, it's at least an older bug.

Issue #399, opened 2018-12-12
date().required() does not validate correctly within when()


pcorey:
Given this code:

```
const { date, mixed, object, string } = require('yup');

const schema = object().shape({
    foo: string(),
    date: mixed().when('foo', {
        is: 'bar',
        then: date().required()
    })
});

console.log(schema.validateSync({ foo: 'boo' }));
console.log(schema.validateSync({ foo: 'bar', date: '123' }));
console.log(schema.validateSync({ foo: 'bar', date: 'hi' }));
```

The first two calls to `schema.validateSync` pass, as expected. The third call that validates `{ foo: 'bar', date: 'hi' }` also passes validation, which seems wrong. `'hi'` is not a valid date, and `date()` wouldn't accept this value outside of a `while()` block.

I'm on `yup@0.26.6`.

jquense:
Seems like a bug!

Issue #381, opened 2018-11-28
Error when use `when` combined with `lazy`


Ridermansb:
I received this error `targetVal.concat is not a function​​
  at ​​​merge​​​ `

```
const creditCardSchema = yup.object().shape({
  addressNumber: yup.lazy(val => {
    switch (typeof val) {
      case "number":
        return yup.number()
      default:
        return yup
          .string()
          .matches(/(S\/N)/i);
    }
  })
});

const paymentSchema = yup.object().shape({
  creditCard: yup.object().when("paymentMethod", {
    is: "credit_card",
    then: creditCardSchema
  }),
  paymentMethod: yup.string().required()
});

const checkoutSchema = yup.object().shape({
  payment: paymentSchema
});

checkoutSchema.validate({
    payment: {
        paymentMethod: "credit_card"
    }
}); // Throw errror targetVal.concat is not a function​​
  at ​​​merge​​​ 
```

You can check online ... 
https://runkit.com/ridermansb/yup-error-when-lazy

PR #444, opened 2019-02-03
fix: concat of mixed and subtype

vonagam:
Separate PR from #443.

```js
mixed().concat(number()).validateSync([]) // should throw
```


vonagam:
By the way, fixes #399 (in issue's example a problem in concating mixed and date, not in when).

vonagam:
Additional commit (can be separated into separate PR, but then i will have to make rebase after this PR will be merged). Small one, but fixes #381.

Smallest repo of a problem:

```js
yup.object({
  conditional: yup.object().when("$check", {
    is: true,
    then: yup.object({
      lazy: yup.lazy(val => yup.number()),
    })
  }),
}).validateSync({}, {context: {check: true}});
```

It fixes it, but underlying cause of the issue still there:

- `validate` in non-strict mode calls `cast`.

- In `object._cast` field get [`resolve`d](https://github.com/jquense/yup/blob/d02ff5e59e004b4c5189d1b9fc0055cff45c61df/src/object.js#L104) then [`cast`ed](https://github.com/jquense/yup/blob/d02ff5e59e004b4c5189d1b9fc0055cff45c61df/src/object.js#L113).

- `cast` [calls](https://github.com/jquense/yup/blob/d02ff5e59e004b4c5189d1b9fc0055cff45c61df/src/mixed.js#L152) `resolve` inside itself. So you end up with `schema.resolve().resolve()`.

- after `resolve` with conditions is executed resulted schema still contains original schema conditions, they are not stripped. So when you call `resolve` on resulted schema, those conditions will be called again and their result will be `concat`ed again.

- and if in a result of conditions was a lazy schema, it ended up `concat`ing with itself, but `Lazy` do not have `concat` method and error was thrown.

Current solution solves the last point: skip merging same schema's or values.

Breaking addition will be a removal of original `_conditions` from `resolve` result. It will also allow to easily implement nested `when`s.

src/util/merge.js
```diff
@@ -1,30 +1,27 @@
...
 
-      if (sourceVal === undefined) continue;
-
-      if (isSchema(sourceVal)) {
-        target[key] = isSchema(targetVal)
-          ? targetVal.concat(sourceVal)
-          : sourceVal;
+      if (sourceVal === undefined) {
+        source[key] = targetVal;
+      } else if (sourceVal === targetVal || sourceVal instanceof Lazy) {
```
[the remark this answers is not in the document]
vonagam: Done.

src/util/merge.js
```diff
@@ -1,30 +1,27 @@
...
       var targetVal = target[key],
         sourceVal = source[key];
 
-      if (sourceVal === undefined) continue;
-
-      if (isSchema(sourceVal)) {
-        target[key] = isSchema(targetVal)
-          ? targetVal.concat(sourceVal)
-          : sourceVal;
+      if (sourceVal === undefined) {
```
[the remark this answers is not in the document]
vonagam: We go through keys of target. So:

```js
if (sourceVal === undefined) {
  source[key] = targetVal;
} else ...
```

Take for example `_default` value. If it is present on `target` but not on `source` we still want to `source[key] = targetVal` even though `targetVal === undefined`.

src/util/merge.js
```diff
@@ -1,30 +1,27 @@
 import has from 'lodash/has';
 import isSchema from './isSchema';
+import Lazy from '../Lazy';
 
 let isObject = obj => Object.prototype.toString.call(obj) === '[object Object]';
 
 export default function merge(target, source) {
-  for (var key in source)
-    if (has(source, key)) {
+  for (var key in target)
```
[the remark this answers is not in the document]
vonagam: Done. Renamed to `prependDeep` and changed order. unresolve If you have better ideas for name.

src/util/merge.js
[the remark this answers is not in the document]
jquense: that's fine, sorry forgot there was was `has` check above

src/mixed.js
```diff
@@ -112,7 +112,7 @@ const proto = (SchemaType.prototype = {
         } and ${schema._type}`,
       );
     var cloned = this.clone();
-    var next = merge(this.clone(), schema.clone());
+    var next = prependDeep(schema.clone(), this.clone());
```
jquense: now that this isn't mutating the first value the second `clone` should be unnecessary, can use `cloned` above or just `this`

src/mixed.js
```diff
@@ -112,7 +112,7 @@ const proto = (SchemaType.prototype = {
...
 
     // undefined isn't merged over, but is a valid value for default
     if (has(schema, '_default')) next._default = schema._default;
```
jquense: this isn't true anymore

vonagam: Still true. If `this` has a non-`undefined` default value and `schema` has default value of `undefined`, then resulted one will have non-`undefined` one, we don't want that, we want to preserve override by `undefined`.

src/mixed.js
vonagam: Right. Now uses `this`, plus `cloned` above not needed too. So 2 clones less.

src/mixed.js
jquense: I meant the comment. `undefined` isn't merged over from `this` unless it's also `undefined` in `schema`. the default value will still get clobbered yes. Tho i wonder if the `undefined` check in `prependDeep` should just be a `has` instead of strict undefined check

vonagam: Updated comment. This one will do? (If not, there is a feature in github where you can add a  suggestion here in comments and i can easily add it to PR)

jquense:
thanks!

PR #445, opened 2019-02-03
feat: use a schema as when condition

vonagam:
Separate PR from #443.

### DIFF
--- a/src/Condition.js
+++ b/src/Condition.js
@@ -7,6 +7,10 @@ function callOrConcat(schema) {
   return base => base.concat(schema);
 }
 
+function makeIsFn(refs, predicate) {
+  return refs.length < 2 ? predicate : (...values) => values.every(predicate);
+}
+
 class Conditional {
   constructor(refs, options) {
     let { is, then, otherwise } = options;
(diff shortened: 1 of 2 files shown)


jquense:
Thanks for sketching this out. I'm not convinced this API is worth the added code, I've gone back and forth a few times on adding this and overall I don't think the caveats and gotchas are probably worth the convenience. The biggest problem is that you need a sync capable schema which is not obvious from the API. And if you do want to use a sync schema it's easy to use one explicitly with the function form of `is`.

Overall I think trying to keep the core library small is probably more valuable here than the API affordance but let me know what you think

vonagam:
So, this change requires just 5 lines of code. 

Change from:

```jjs
let isFn =
  typeof is === 'function'	
    ? is
    : (...values) => values.every(value => value === is);
```

to:

```
let isFn;

if (typeof is === 'function') {
  isFn = is;
} else if (isSchema(is)) {
  isFn = (...values) => values.every(value => is.isValidSync(value));
} else {
  isFn = (...values) => values.every(value => value === is);
}
```

Plus i added here `makeIsFn` function so that we do not create and destroy arrays when we deal with only one value (i assume it is the most popular case). Optional 4 lines.

And the only api change here is that you can now use a sync schema as a condition.

So, i do not see here big costs nor in code increase, nor in api complication. 

As for benefits, i think that it is expected that you can use validation schema as a condition (like, they are made for this) without writing custom helper. Plus joi provides such usage, and i see gain in providing same functionality for same named methods (less code change for a user when switching and all joi schemas are sync).

But yes, it does not provide functionality which you currently cannot do. So, while i prefer not to write custom helper for this in my codebase, i am ok with this PR not being merged. Feel free to close without any comments.

vonagam:
Possible bonus point for this: if `is`, `then` and `otherwise` are schemas then maybe we can use this fact later when implementing `describe` for schema conditions. (If there will be any way to get `describe()` without finalisation)

vonagam:
This pull should wait for #447.

So that `is.isValidSync(value)` could be easily replaced with `is.isValidSync(value, options)`.

PR #446, opened 2019-02-03
fix: remove README js inconsistencies

vonagam:
Remove some inconsistencies in readme js snippets.

Most changes:

- let or const to var

- `inst` to `schema`

- \`\`\`javascript to \`\`\`js

- // -> or //=> to // =>

- await for async calls

- space after comment and first letter is small

- blank line between schema creation and usage

- bool() to boolean()

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,14 +122,14 @@ When using es module syntax, yup exports everything as a named export
 
 ```js
 import * as yup from 'yup'; // for everything
-// or only what you need
-import { string, object } from 'yup';
+// or
+import { string, object } from 'yup'; // for only what you need
 ```
 
 ```js
-var yup = require('yup');
+let yup = require('yup');
 
-var schema = yup.object().shape({
+let schema = yup.object().shape({
   name: yup.string().required(),
   age: yup
     .number()
@@ -143,7 +143,7 @@ var schema = yup.object().shape({
   }),
 });
 
-//check validity
+// check validity
 schema
   .isValid({
     name: 'jimmy',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -181,13 +181,14 @@ setLocale({
   },
 });
 
-// Now use Yup schemas AFTER you defined your custom dictionary
-const schema = yup.object().shape({
+// now use Yup schemas AFTER you defined your custom dictionary
+var schema = yup.object().shape({
```
jquense: Let's leave the const and let calls

README.md
```diff
@@ -424,10 +429,10 @@ but uses the resulting schema as the subject for validation.
 > Note! The `value` here is the _root_ value relative to the starting schema, not the value at the nested path.
 
 ```js
-const schema = object({
+var schema = object({
   foo: array().of(
     object({
-      loose: bool(),
+      loose: boolean(),
```
jquense: I generally perfer bool since it's not a reserved word

README.md
vonagam: Do you say just to leave this lines or to change all `var`s to `let` or `const`. 
Personally i prefer second option with `let`, i chose `var` as it was most used here before. 
If you mean just to leave, then is there any rule or a reason?

README.md
vonagam: But `boolean` is not reserved word too, like `string`, `number` or others. Am i missing something?

In api section `boolean` is on the same levels as others: having section named `boolean` and listed in exports, mentioning only in comments that there is also `bool` alias.

README.md
jquense: Sorry yeah, I would perfer let everywhere 👍

vonagam: Done

Issue #335, opened 2018-10-08
is it possible to have nested .whens()?


TiagoFuelber:
I'm trying to achieve nested .when() calls, but with no success.

I'm trying something similiar to the following example:

```
const schema = Yup.object().shape({
  day: Yup.number(),
  month: Yup.number(),
  year: Yup.number().when("day", {
    is: 1,
    then: Yup.number.when("month", {
      is: 1,
      then: Yup.number().required("Its january, please select the year"),
      otherwise: Yup.number().required(
        "Its not january, please select the year"
      )
    }),
    otherwise: Yup.number()
  })
});
```

Does anyone have an idea of how can I achieve that?

jquense:
Nested `when`s don't work unfortunately, but you can flatten them by doing a when for multiple keys at once

``` 
when(['day', 'month'], { is: (day, month) => true, ... })
```

ugeng:
Hi, @jquense. Is it possible at all? And how it could be achieved?

jquense:
@ugeng at the moment no there isn't any support for it. There isn't a strong reason why, but it's complicated enough that it's not super simple to add. I'm happy to evaluate a PR tho if someone is up for it.

ugeng:
Well, the strong reasons are purity, readability and testability of code...
I wish I have more free time for that... 
At the moment, conditional logic in my schemes is not way too complicated (and I really can "flatten" predicates seamlessly), but when it does I would gladly (or sadly) dive deeper in this feature..
Thanks.

vinicius5581:
While this doesn’t exist yet, have anyone found a work around? 

I need something like this:


const schematic = Yup.object().shape({
       aBoolean: Yup.bookean(),
       aThing: Yup.object().when(“aBoolean”, {
              is: true,
              then: object().shape({
                    aString: Yup.string(),
                    anotherString: Yup.string().when(“aString”, {
                           is: “theRightThing”,
                           then: Yup.string().required()
                     })
              }).required()

sibelius:
@jquense any plans to support nested whens?

darrensw777:
As of quarantine 2020, this now works as you think it should:

`lastDancingDay: Yup.string().when('myDance', {
  is: true,
  then: Yup.string().when('doYouTango', {
    is: 'Yes',
    then: Yup.string().required('Dancing shoes required'),
  }),
}),`

wasssim1:
I think the .nullable() is still not handled in nested "whens"

PR #447, opened 2019-02-04
feat: finalize resolve()

### LINKED ISSUE #381: Error when use `when` combined with `lazy`

vonagam:
This PR allows safely nest `lazy`s and `when`s in themselves and each other. (Asked in #335 for example)

Currently there is a difference between calling `resolve()` on `lazy` and on `mixed` with `when`s:

- `lazy.resolve()` will return a schema not connected to `lazy` instance. So calling `resolve()` on a returned schema does not call `lazy` body anymore.

- `mixed.resolve()` will return a schema resulted by `concat`ing base schema and conditional one, as nor `mixed.concat()` nor `mixed.resolve()` remove initial `when` conditions they are present in resulted schema and those conditions get called and merge their results on subsequent `resolve` calls of merged schema.

So, in a second case it is not a final resolve, but a partial one (resulted schema will contain conditions). For most cases it does not matter (though it cases unnecessary calls and may cause bugs for example with `reach`).

Here all `resolve`s are made equal and expect to return a non-lazy schema without any conditions remaining. So calling `resolve()` on a resulted schema will simply return it back. This also allows to make nested conditions and lazy to work without a problem.

It is a breaking change though, since before `reach()`ed schema's with `when`s where only partially resolved and expected to run those conditions again. This should be addressed. I think that there are three possible ways:

- Do not resolve `reach`ed schema. And if people need some methods which require resolved schema (like `meta()`) tell them about `resolve()`. So making `resolve()` part of public api.

- Keep `reach` behaviour the same, but add new function `getAt` (or something) which will return non resolved schema. And to tell people to use it now for most cases.

- Modify signature of `reach`, introduce flag for resolving or not. (Plus a `parent` option, since it may be used in top level `resolve` call.) So maybe reach will become `reach(schema, {path, value, parent, context, resolve})`.

(Personally i think first one is better, maybe third. Just suggestions.)

Half of proposal is to standardise `resolve()` behaviour and other half is to use created opportunity to nest things freely. 

(This also fixes #381 without #444, as now subsequent `resolve()` calls to do not cause repeats of `concat`s)

What do you think?

### DIFF
--- a/src/Condition.js
+++ b/src/Condition.js
@@ -3,8 +3,8 @@ import isSchema from './util/isSchema';
 
 function callOrConcat(schema) {
   if (typeof schema === 'function') return schema;
-
-  return base => base.concat(schema);
+  if (!schema) return base => base;
+  return (base, options) => base.concat(schema.resolve(options));
 }
 
 class Conditional {
@@ -32,10 +32,11 @@ class Conditional {
           : (...values) => values.every(value => value === is);
 
       this.fn = function(...values) {
-        let currentSchema = values.pop();
+        let options = values.pop();
+        let schema = values.pop();
         let option = isFn(...values) ? then : otherwise;
 
-        return option(currentSchema);
+        return option(schema, options);
       };
     }
   }
@@ -43,12 +44,14 @@ class Conditional {
   resolve(ctx, options) {
     let values = this.refs.map(ref => ref.getValue(options));
 
-    let schema = this.fn.apply(ctx, values.concat(ctx));
+    let schema = this.fn.apply(ctx, values.concat(ctx, options));
 
-    if (schema !== undefined && !isSchema(schema))
+    if (schema === undefined) return ctx;
+
+    if (!isSchema(schema))
       throw new TypeError('conditions must return a schema object');
 
-    return schema || ctx;
+    return schema.resolve(options);
   }
 }
 
--- a/src/Condition.js
+++ b/src/Condition.js
@@ -1,52 +1,58 @@
 import has from 'lodash/has';
 import isSchema from './util/isSchema';
 
-function callOrConcat(schema) {
-  if (typeof schema === 'function') return schema;
-  if (!schema) return base => base;
-  return (base, options) => base.concat(schema.resolve(options));
+function wrapCusomFn(fn) {
+  return function(...args) {
+    args.pop();
+    return fn.apply(this, args);
+  };
 }
 
-class Conditional {
-  constructor(refs, options) {
-    let { is, then, otherwise } = options;
+function makeFn(options) {
+  if (typeof options === 'function') return wrapCusomFn(options);
 
-    this.refs = [].concat(refs);
+  if (!has(options, 'is'))
+    throw new TypeError('`is:` is required for `when()` conditions');
 
-    then = callOrConcat(then);
-    otherwise = callOrConcat(otherwise);
+  if (!options.then && !options.otherwise)
+    throw new TypeError(
+      'either `then:` or `otherwise:` is required for `when()` conditions',
+    );
 
-    if (typeof options === 'function') this.fn = options;
-    else {
-      if (!has(options, 'is'))
-        throw new TypeError('`is:` is required for `when()` conditions');
+  let { is, then, otherwise } = options;
 
-      if (!options.then && !options.otherwise)
-        throw new TypeError(
-          'either `then:` or `otherwise:` is required for `when()` conditions',
-        );
-
-      let isFn =
-        typeof is === 'function'
-          ? is
-          : (...values) => values.every(value => value === is);
-
-      this.fn = function(...values) {
-        let options = values.pop();
-        let schema = values.pop();
-        let option = isFn(...values) ? then : otherwise;
-
-        return option(schema, options);
-      };
-    }
+  let check;
+  if (typeof is === 'function') {
+    check = is;
+  } else {
+    check = (...values) => values.every(value => value === is);
   }
 
-  resolve(ctx, options) {
+  let fn = function(...args) {
+    let options = args.pop();
+    let schema = args.pop();
+    let branch = check(...args) ? then : otherwise;
+
+    if (!branch) return undefined;
+    if (typeof branch === 'function') return branch(schema);
+    return schema.concat(branch.resolve(options));
+  };
+
+  return fn;
+}
+
+class Condition {
+  constructor(refs, options) {
+    this.refs = refs;
+    this.fn = makeFn(options);
+  }
+
+  resolve(base, options) {
     let values = this.refs.map(ref => ref.getValue(options));
 
-    let schema = this.fn.apply(ctx, values.concat(ctx, options));
+    let schema = this.fn.apply(base, values.concat(base, options));
 
-    if (schema === undefined) return ctx;
+    if (schema === undefined || schema === base) return base;
 
     if (!isSchema(schema))
       throw new TypeError('conditions must return a schema object');
@@ -55,4 +61,4 @@ class Conditional {
   }
 }
 
-export default Conditional;
+export default Condition;
--- a/src/Lazy.js
+++ b/src/Lazy.js
@@ -2,16 +2,17 @@ import isSchema from './util/isSchema';
 
 class Lazy {
   constructor(mapFn) {
-    this._resolve = (...args) => {
-      let schema = mapFn(...args);
+    this._resolve = (value, options) => {
+      let schema = mapFn(value, options);
+
       if (!isSchema(schema))
         throw new TypeError('lazy() functions must return a valid schema');
 
-      return schema;
+      return schema.resolve(options);
     };
   }
-  resolve({ value, ...rest }) {
-    return this._resolve(value, rest);
+  resolve(options) {
+    return this._resolve(options.value, options);
   }
   cast(value, options) {
     return this._resolve(value, options).cast(value, options);
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -135,14 +135,22 @@ const proto = (SchemaType.prototype = {
   },
 
   resolve(options) {
-    if (this._conditions.length) {
-      return this._conditions.reduce(
+    let schema = this;
+
+    if (schema._conditions.length) {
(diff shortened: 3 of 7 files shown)


jquense:
Hey thanks again for taking the time to dig into these hairier issues. I'd like to take a step back on this one and frame the problem we are trying to solve at a higher level. 

Right now you can't reasonably nest `lazy` or another `when` in a when, and vice versa. I do think this is annoying and not obvious. I've always ment to explicitly document that it's not supported. I would counter a bit by saying that nesting is not strictly necessarily. Anything (most?) can be flattened into a single a `when` or lone `lazy` call. In my experience using yup that's usually fine, but that would depend on how small and portable one makes schema and app workflows etc. 

In terms of `reach` tradeoffs I think that's a big concern. Form libraries depend on the resolving nature of `reach`, resolved schema metadata to make form choices, etc. On a functional level tho `reach` _must_ resolve schema b/c a path may not exist until the schema is resolved.

We would let people deal with that nuance themselves but I would consider users needing to think about `resolve` explicitly an API failure in terms of mental overhead and complexity.

In terms of implementation I'd love to make `resolve` more consistent, It was added kinda adhoc with Lazy to try and unify the schema interface so most schema don't need to internally care about whether something is lazy or not. 

> mixed.resolve() will return a schema resulted by concating base schema and conditional one, as nor mixed.concat() nor mixed.resolve() remove initial when conditions they are present in resulted schema and those conditions get called and merge their results on subsequent resolve calls of merged schema.

Yeah this is annoying, and kinda a result of the overly generalized approach to schema implementation yup uses. I'd love to fix this, I'm just a bit concerned about the implementations I need to dig into how it affects `reach` and the like, The _intent_ was that resolve finalizes, so there should be a way to deal with this that still handles edge cases.

jquense:
quick specific thoughts: 

- we only need to resolve up to the final returned schema from `reach` 

**reach use -cases
- validating or casting at a path: `reach('blah', ...).validate()`, which is now mostly covered by `validateAt()` and friends (which use reach). no need to finalize since `validate` does right after but doesn't hurt if you do finalize it.

- Using a nested schema elsewhere `object({ foo: reach(other, 'blah') })` I think this is better served by just exporting the nested schema where it's defined, but meh. it's unclear whether this should fully resolve or not, i lean towards it should.

- schema introspection, i.e. checking meta/describe, etc on a nested path. this should finalize the entire chain.

jquense:
> reach(schema, 'blah', valueA, contextA).validate(valueB, {context: contextB});

Ideally i think this shouldn't be something they have to think about. again i think not finalizing in that case is the preferable one, it's not not terrible if we do (assuming ppl pas the same context both places which is the common case)


> i expect it to stay the same and react to changes (and especially not to finalise on undefined value and context).

I think that would weird depending on the situation, but generally conditions are specific to a broader context, which would have changed if you use it elsewhere. e.g. the target location of the schema may not have the same parent structure or siblings.

```js
const Foo = object({
  other: bool(),
  name: string().when('other', { is: true, then: s => s.required() )
})

const Bar = object({
 name: reach(Foo, 'name', options)
})
```

I think it would be confusing that name still depended on 'other' existing in the above


> Option: maybe a returned schema from reach will not be finalised but will store options with which it was reached and if it needs to be finalised for a call (like describe), it can use those options at that moment. Sounds like a promising direction for me.

That may be a good direction, tho i'm not excited about the idea of schema being stateful in this way, it may end up complicating everything over all in terms of ease of use and understanding.


I think it's fine to start with _not_ resolving the final return value from `reach` for now, it's always possible to resolve a schema, less so to "unresolve" it.

vonagam:
> I think it's fine to start with not resolving the final return value from reach for now, it's always possible to resolve a schema, less so to "unresolve" it.

I am for not resolving result. That would be a better solution, though a breaking one. Proposal for storing `reach`ing options was made with assumption that you do not want breaking changes yet.

What about `describe`? Should we add second optional argument for resolving?

And for future: what about `parent` resolving option? It is not documented in readme for being part of `validate` [options](https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror) and not present in `reach` signature. Is it part of public api or is it considered internal only right now? I assume it is needed option if you use `reach` to get element with possible refs to siblings.

jquense:
> [...] assumption that you do not want breaking changes yet.

no, i'm fine with breaking changes as long as we preserve required use-cases.

> What about describe? Should we add second optional argument for resolving?

Let's punt on that for right now. this is already inconsistent depending on if you call it after `reach`ing or just on the schema, i'm ok with it being sub optimal but consistent for the near-term.

> And for future: what about parent resolving option?

Yeah it's for internal usage right now, which is why it's not documented. We can try and keep it private until it doesn't make sense anymore

test/yup.js
```diff
@@ -125,21 +125,24 @@ describe('Yup', function() {
       },
     };
 
-    reach(inst, 'nested.arr.num', value).should.equal(num);
-    reach(inst, 'nested.arr[].num', value).should.equal(num);
+    // reach(inst, 'nested.arr.num', value).should.equal(num);
+    // reach(inst, 'nested.arr[].num', value).should.equal(num);
```
jquense: lets add these back but with the needed `resolve()` call

src/Condition.js
```diff
@@ -32,27 +32,28 @@ class Conditional {
           : (...values) => values.every(value => value === is);
 
       this.fn = function(...values) {
-        let currentSchema = values.pop();
+        let options = values.pop();
+        let schema = values.pop();
         let option = isFn(...values) ? then : otherwise;
 
-        return option(currentSchema);
+        return option(schema, options);
```
jquense: why the API change here?

test/object.js
```diff
@@ -437,8 +437,8 @@ describe('Object types', () => {
         }),
       });
 
-      reach(inst, 'nested').should.equal(inst);
-      reach(inst, 'x.y').should.equal(inst);
+      reach(inst, 'nested').should.equal(inst.fields.nested);
```
jquense: was this test just broken?

src/mixed.js
```diff
@@ -136,16 +136,21 @@ const proto = (SchemaType.prototype = {
...
-  resolve({ context, parent }) {
-    if (this._conditions.length) {
-      return this._conditions.reduce(
-        (schema, match) =>
-          match.resolve(schema, match.getValue(parent, context)),
-        this,
+  resolve(options) {
+    let schema = this;
+
+    while (schema._conditions.length) {
```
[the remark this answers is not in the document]
vonagam: Replaced with recursion, will throw on stack exceeding. (Plus may be better for debugging)

src/Condition.js
vonagam: First argument is the same. Additional second argument is used in this:

```js
return (base, options) => base.concat(schema.resolve(options));
```

So that `Lazy` can be `concat`ed.

test/yup.js
vonagam: Done.

src/Lazy.js
```diff
@@ -2,16 +2,17 @@ import isSchema from './util/isSchema';
...
         throw new TypeError('lazy() functions must return a valid schema');
 
-      return schema;
+      return schema.resolve(options);
     };
   }
-  resolve({ value, ...rest }) {
-    return this._resolve(value, rest);
+  resolve(options) {
+    return this._resolve(options.value, options);
```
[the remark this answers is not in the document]
vonagam: To allow nested `Lazy` or `when` to work we need not remove anything from `options`, so that you can `schema.resolve(options)` on returned schema from `lazy` function.

test/object.js
vonagam: Added `resolve()` as in `tests/yup`.

src/Condition.js
jquense: I see, let's inline the `callOrConcat` function here, it's not adding any value separate. I'm incline to say we should keep the public api the same for when it's a function...

src/Lazy.js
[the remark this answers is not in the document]
jquense: hmm yeah, that's not great, it also means that every schema needs to now treat `options` as dynamic object bag instead of a defined shape that applies to all schema...it's fine for now. I may rethink how this is all structured, there are to many divergent paths between `lazy` and `when`

src/Condition.js
vonagam: Did rebase on current master and did refactor Condition.js file. Just split logic into separate functions, no functional changes except requested removal of `options` object from custom functions api.

src/Condition.js
```diff
@@ -1,59 +1,75 @@
...
-        let option = isFn(...values) ? then : otherwise;
+  let { is, then, otherwise } = options;
 
-        return option(currentSchema);
-      };
-    }
-  }
+  is = makeConditionPredicate(is);
+  then = makeBranchCustomizer(then);
+  otherwise = makeBranchCustomizer(otherwise);
```
jquense: again lets remove `makeBranchCustomizer` (as well as `makeConditionPredicate`) and move it inline to line 47

vonagam: Any reason? That way it is more performant.

jquense: its mild indirection for an unverified perf benefit. I'm skeptical even that it could/would be, the above creates a bunch of unneeded closures and functions.

vonagam: Done.

Issue #418, opened 2019-01-10
String .length() method not documented in the README.


DaddyWarbucks:
The `.length()` method found here: https://github.com/jquense/yup/blob/6ab83bcc1c0bdf4febb237c4961d6e48ffd201ec/src/string.js#L40
is not documented in the README.

I will be happy to submit a PR for this, just documenting it here first.

jquense:
ya happy to take a PR 👍

DaddyWarbucks:
Solved and closed by [PR #448 ](https://github.com/jquense/yup/pull/448)

PR #448, opened 2019-02-05
Document .string().length() method

DaddyWarbucks:
Solves [PR #418](https://github.com/jquense/yup/issues/418) by documenting the `yup.string().length()` method in the README. I also noticed a few misuses of `a` and `an` and cleaned those up.


jquense:
Thanks:

PR #454, opened 2019-02-09
feat: various improvements and fixes

vonagam:
### Fixes

1) **Fixes** `array.notRequired()`/`string.notRequired()`. Currently they have overridden  `required()`, but not `notRequired()`.

```js
array().required().notRequired().isValidSync([]) // now true, not false
```

2) **Fixes** `cast()` for `object()` and value with a key which is present in `Object.prototype`.

```js
object().required().cast({toString: 5}); // now does not throw
```

3) **Fixes** `withMutation()` and nested invocation.

```js
mixed().withMutation(schema => {
  schema.withMutation(changeA);
  changeB(schema);
}); // now changeB gets mutable instance
```

4) **Fixes** `describe()` which currently ignores tests with same name but different params, showing only first one of them. 

```js
string().matches(/a/).matches(/b/).describe() // now have two tests, not one
```

### Additions

5) **Adds** `skipAbsent` option to tests. Currently most tests in codebase have `isAbsent(value)` check in their functions. To skip those tests from the start and make functions cleaner, the option was added.

```js
this.test({
  skipAbsent: true,
  ...
}); // this test will be skipped if a value is null or undefined
```

6) **Adds** boolean argument to `strict()` and `nullable()` in case when one wants to revert effect. (`strict()` has such argument mentioned in readme, but currently it is not present)

```js
mixed().nullable().nullable(false); // not nullable anymore
```

7) **Adds** a function argument to `clone()`, if present is passed to `withMutation()` of a clone. Uses this to reduce cloning in some method.

```js
this.clone(next => {
  next.transform(...);
  next.test(...);
}); // instead of this.transform().test() which does cloning twice
```

8) **Adds** `optional()` alias for `notRequired()` and `unknown()` as opposite to `noUnknown()`.

```is
number().optional(); // reads better and present in joi, same with unknown().
```

### Changes

9) **Changes (fixes)** behaviour of `notRequired()` not concating and `noUnknown` still being present in `describe()` output even if `noAllow` is `false`. Both problems connected with way to remove a test. `notRequired()` simply removes test by name, but this leaves no trace of user's action or intent, so it is lost for `concat()`. `noUnknown()` uses `exclusive` option to replace test with new one, but it needlessly runs and gets present in `describe`. Solution is to add new test option `skip`. Tests with `skip` as `true` do not run and do not get `describe`d. Useful with `exclusive` option. Plus added `mixed.skip(name)` method, which adds a test with provided `name` and `exclusive`/`skip` as `true`.

```js
object().noUnknown().concat(object.noUnknown(false)).validateSync({ foo: 'bar' });
// => true, as before

mixed().required().concat(mixed()).validateSync(undefined);
// => false, as before

mixed().required().concat(mixed().notRequired()).validateSync(undefined);
// => true, not false as before
```

Will split into multiple PR later. Opened to get feedback. I assume there will be no problems with first seven ones. 

- Do you have any problems with adding aliases in number 8?
- Should last example in number 9 return `true` or `false`? If `false`, why?

### DIFF
--- a/src/array.js
+++ b/src/array.js
@@ -1,17 +1,12 @@
 import inherits from './util/inherits';
-import isAbsent from './util/isAbsent';
 import isSchema from './util/isSchema';
 import makePath from './util/makePath';
 import printValue from './util/printValue';
 import MixedSchema from './mixed';
-import { mixed, array as locale } from './locale';
+import { array as locale } from './locale';
 import runValidations, { propagateErrors } from './util/runValidations';
 
-let hasLength = value => !isAbsent(value) && value.length > 0;
-
-export default ArraySchema;
-
-function ArraySchema(type) {
+export default function ArraySchema(type) {
   if (!(this instanceof ArraySchema)) return new ArraySchema(type);
 
   MixedSchema.call(this, { type: 'array' });
@@ -94,9 +89,7 @@ inherits(ArraySchema, MixedSchema, {
             originalValue: originalValue[idx],
           };
 
-          if (subType.validate) return subType.validate(item, innerOptions);
-
-          return true;
+          return subType.validate(item, innerOptions);
         });
 
         return runValidations({
@@ -110,9 +103,11 @@ inherits(ArraySchema, MixedSchema, {
       });
   },
 
-  of(schema) {
-    var next = this.clone();
+  _isFilled(value) {
+    return value.length > 0;
+  },
 
+  of(schema) {
     if (schema !== false && !isSchema(schema))
       throw new TypeError(
         '`array.of()` sub-schema must be a valid yup schema, or `false` to negate a current sub-schema. ' +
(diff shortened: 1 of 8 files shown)


vonagam:
**Update**:

1) `string.required()`/`array.required()` **opened PR, code present**.

2) `object()` and values from prototype **opened PR, code removed**.

3) Nested `withMutation()` **opened PR, code present**.

4) `describe()` and name uniqueness  - no change (no point in removing, changes same lines as `skip`).

5) `skipAbsent` - no change.

6) `strict(false)` **opened PR, code removed**.

7) `clone(fn)` **opened PR, code present**.

8) aliases `optional`, `unknown` - **opened PR, code removed**.

9) `skip` - no change.

So now this pull is about `skip`ing stuff and small code refactors.

PR #455, opened 2019-02-10
fix: object and prototype keys

vonagam:
Number 2 from #454

`fields = Object.create(null)` does not work as a way to ignore prototype properties since after first `clone()` it will be a usual object.

```js
object().required().validateSync({toString: 5}); // now does not throw
```

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -38,7 +38,7 @@ export default function ObjectSchema(spec) {
     },
   });
 
-  this.fields = Object.create(null);
+  this.fields = {};
   this._nodes = [];
   this._excludedEdges = [];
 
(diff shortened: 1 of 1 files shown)


jquense:
Hmm that seems like wrong behavior in lodash's clone, and a bit concerning that its not preserving the prototype of objects on copy

vonagam:
Well, i think it is safer to expect and treat everything as a simple object. You have checks for `fields` everywhere: `has` used for this in `sortFields` and `validate()` has check for presence of `field.validate`, only in this place check for missing.

Another way is to allow subtype to customise `cloneDeepWith` customizer used in `clone()`. But it doesn't worth it just for this issue.

PR #456, opened 2019-02-10
feat: allow withMutation() nesting

vonagam:
Number 3 from #454

```js
mixed().withMutation(schema => {
  schema.withMutation(changeA);
  changeB(schema);
}); // now changeB gets mutable instance
```

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -96,9 +96,10 @@ const proto = (SchemaType.prototype = {
   },
 
   withMutation(fn) {
+    let before = this._mutate;
     this._mutate = true;
     let result = fn(this);
-    this._mutate = false;
+    this._mutate = before;
     return result;
   },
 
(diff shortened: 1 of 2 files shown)


jquense:
What would anyone need to do this?

vonagam:
Right now `withMutation()` limited only to constructor, but if you want to use it anywhere else in codebase to reduce number of cloning (deep clones with lodash are not cheep) you need to be sure that it can be nested:

```js
schema.withMutation(schema => {
  schema.methodA(); // if methodA uses withMutation()
  schema.test({...}); // then this will have no effect on schema
});
```

vonagam:
`withMutation()` is part of public api as it is documented in readme. And inability to use it inside itself is not mentioned. So this is kinda bug, which is fixed by one line.

jquense:
I agree it's technically a bug, still worth understanding the use case since i've got to maintain it. I don't think the above is very convincing as a use-case but sure.

PR #458, opened 2019-02-10
feat: clone tapping

vonagam:
Number 7 from #454

```js
this.clone(next => {
  next.transform(...);
  next.test(...);
}); // instead of this.transform().test() which does cloning twice
```


jquense:
Hmm that seems like wrong behavior in lodash's clone, and a bit concerning that its not preserving the prototype of objects on copy

PR #462, opened 2019-02-10
fix: remove unneeded clone() in describe()

vonagam:
What's the purpose of clone here?

If you still want to make clone (to prevent modification for `_meta` and `OPTIONS.params`), we can simply wrap the result in `cloneDeepWith()`.

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -480,13 +480,11 @@ const proto = (SchemaType.prototype = {
   },
 
   describe() {
-    const next = this.clone();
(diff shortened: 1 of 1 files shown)


jquense:
yes to prevent accidental mutation of instance properties. `clone` is already called a lot just in normal schema usage, if it was a significant bottleneck optimizing here by reducing the stuff cloned wouldn't really solve the problem, lets not over optimize without good cause 👍

PR #469, opened 2019-02-21
enhancement(readme): add zero inclusive comments

Fajer39:
We had a confusion in our project about whether `number().positive()` or `number().negative()` includes zero or not, thus this little enhancement in the README would come handy for other like-minded folks. :)


jquense:
see #417 and #322 I'd prefer we fix the methods instead to not include zero 👍

Fajer39:
@jquense I see, that's for sure a better approach :) Closing then.

PR #480, opened 2019-02-27
Added precision method on number

daytonlowell:
Perhaps `decimalSeparator` shouldn't be a parameter on the `precision` method but should instead be part of a locale configuration?


jquense:
I there, thanks for the PR! This feature has been discussed before and we've landed on not adding this to the core library. The main reason is that it's locale and culture dependent and doing that correctly involves specific data which we don't want to be in the business of providing or requiring. It's the same reason the only date format supported for parsing is an ISO format.

PR #488, opened 2019-03-07
change @babel/runtime version to be a range

### LINKED ISSUE #486: @babel/runtime cannot be latest version ?
This actually more of a question than an issue.

I'm trying to reduce the size of my bundle using inspectpack. I noticed that @babel/runtime is duplicated because of yup.
The package.json file indicates a specific version of @babel/runtime:
```json
"@babel/runtime": "7.0.0",
```

My question is: why not `^7.0.0` ? Like the other dependencies. Is there a problem with the latest version of @babel/runtime ?

bsotodo:
Fix #486

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "sinon-chai": "^3.2.0"
   },
   "dependencies": {
-    "@babel/runtime": "7.0.0",
+    "@babel/runtime": "^7.0.0",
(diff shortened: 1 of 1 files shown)


jquense:
thanks

azizhk:
Hi, sorry for pushing, but please can we have a new version published to npm?

PR #491, opened 2019-03-07
Put IntelliJ-like config files on gitignore

Philipp91:
And GitHub is auto-normalizing the line endings.


jquense:
hey there, thanks. However generally its preferable to ignore DIE configs via your own global gitignore https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

Philipp91:
Makes sense, sorry for the noise.

PR #498, opened 2019-03-14
test(when): add test for nested when

sibelius:
test case for #335

not really sure how to test it, as validate will throw

not sure how to fix it either


jquense:
https://github.com/jquense/yup/pull/447

Issue #386, opened 2018-11-30
More strict email validation


victordidenko:
We use yup for client-side validation, and [class-validator](https://github.com/typestack/class-validator) for server-side validation. Class-validator uses [validator](https://github.com/chriso/validator.js) for email validation.
And in some cases email passes yup validation, but fails on server side.

For example `13malinator@wwe.j`

Is it possible to make email validation more strict?
Or should I create issue for validator.js, because email above, frankly speaking, looks like valid according to RFC? But from real world's point of view, this is obviously incorrect email.

inpendio:
Yep, same problem here...

jquense:
you can use whatever regex you'd like with a custom test to match any server logic. I will say, in all likelihood we'll either remove or make the existing test **less strict** in the future, since validating email via a regex is very fraught. You're example above, for instance is definitely a valid email address technically. 

Overall the only real way to validate an email, is to send an email to it and require the user respond.

victordidenko:
Thing is that [validator](https://github.com/chriso/validator.js/blob/master/src/lib/isEmail.js) doesn't use single regex to test email, so it is impossible to write one to match logic...

jquense:
you can use `isEmail` itself in a custom test if you want

schmod:
One glaring omission (which QA folks are incidentally really good at exposing) is that Yup allows email addresses with more than 64 characters before the `@`.

I think we might be able to bolt that onto the regex that Yup uses without meaningfully increasing the size of the library...

ahce:
@jquense why closed this?
I started use yup and get this problem.
I think the same as this [comment](https://github.com/jquense/yup/issues/386#issuecomment-456514231)

PR #508, opened 2019-03-29
fix #507

### LINKED ISSUE #507: yup.string().email() allows invalid email format.
## Summary
The validation for `yup.string().email()` allows an invalid email such as `user@example.com..au` to be marked as valid.

Yup version `0.26.10`

reaysawa:
## Motivation
As told by https://en.wikipedia.org/wiki/Email_address#Local-part, double dots are not allowed either in the local-part or domain.

The change in the regular expression unscopes the dot out of the quantified group so it can only appear at most once after a sequence of valid characters.

This PR also adds some more tests to ensure the current functionality works as expected after the changes.

## Changes
Add more tests for the e-mail format
Fix a case where double dots were allowed in the local-part or domain.


test/string.js
```diff
@@ -176,6 +176,47 @@ describe('String types', () => {
...
+      'user.name+tag+sorting@example.com',
+      'x@example.com',
+      'example-indeed@strange-example.com',
+      'admin@mailserver1',
+      'example@s.example',
+    ].map(email =>
+      v
+        .isValid(email)
+        .should.eventually()
+        .equal(true),
```
reaysawa: @jquense seems like the pre-commit hook added an extra comma there, and I'm not sure if there are problems with that

my original code looked like this

```
    [
      'simple@example.com',
      'very.common@example.com',
      'disposable.style.email.with+symbol@example.com',
      'other.email-with-hyphen@example.com',
      'fully-qualified-domain@example.com',
      'user.name+tag+sorting@example.com',
      'x@example.com',
      'example-indeed@strange-example.com',
      'admin@mailserver1',
      'example@s.example'
    ].map(email => v.isValid(email).should.eventually().equal(true));
```

feel free to push adjustments anywhere if needed

jquense:
Thanks for the PR! I'm a bit on the fence here on one hand this is a simple small change that has better behavior than the current. On the other tho my thinking around an email validation has changed somewhat (see: https://github.com/jquense/yup/issues/386#issuecomment-450428912). Overall I think doing more and more complex regexes on an email is a losing game, the right way to validate an email generally is to send an email to it and have the user respond, which in most web apps is actually the thing that one should do. For more specific or narrowly scoped cases a regex may be fine but that's usually specific to your app's requirements

reaysawa:
@jquense I see. Like you say, it probably has better behavior than the current, so it can be more helpful to eagerly give feedback on those fast validating forms if the user makes a mistake before submitting. It's not supposed to be a true validator, because true validation only happens at the back-end, but it can be helpful in filtering out simple typos.

Since you feel like that's a case not worthy of being considered, I'll close the PR.

Issue #553, opened 2019-06-17
Date ref not does not appear to be getting cast


akmjenkins:
I've got some utility functions for yup that I use like so:

```
export const after = (value, message, schema = yup.date()) => schema.clone().min(value, message);
```

This works:
```
after('2010-01-01', message).isValidSync('2012-01-01') // true
```

This does not:

```
yup.object({
    myField: afterTwo(yup.ref('field2'), message),
}).isValidSync({
    myField:'2012-01-01',
    field2:'2010-01-01'
})
```

From what I can tell in the test, it looks like it resolves the ref correctly, but it doesn't cast it (at least not in the way I think it should cast it, but I could be misunderstanding something).

I can fix it by doing this, (but it's much less elegant :( ):

```
export const after = (value, message, schema = yup.date()) => {
  return schema.clone().test({
    name: 'after',
    message,
    test: function(against) {
      return this.resolve(against) > schema.cast(this.resolve(value));
    },
  });
};
```

I'm just wondering if this is a bug that the resolved ref isn't getting cast in the first example I posted? Am I just misunderstanding how casting is supposed to work (completely possible)? Is it at all related to this old issue: https://github.com/jquense/yup/issues/40?

jquense:
On the face of it, it would make sense it doesn't cast, b/c there is no schema definition for `field2` to use to cast it. Agree the current behavior isn't all that helpful tho. 

also
```
schema.clone().min(value, message);
```
you don't need the `clone()` in your helper, any action of a schema will clone automatically b/c they are immutable

akmjenkins:
@jquense Thanks for the tip! Would it make sense to pass a schema or cast function as an option to the ref? Right now, ref's only take contextPrefix as an option, I could envision an api like this:

```
yup.object({
    myField: afterTwo(yup.ref('field2',{cast:yup.date().cast}), message),
}).isValidSync({
    myField:'2012-01-01',
    field2:'2010-01-01'
})
```

Not sure if it would help to define the whole schema for it like:

```
yup.object({
    myField: afterTwo(yup.ref('field2',{schema:yup.date()}), message),
}).isValidSync({
    myField:'2012-01-01',
    field2:'2010-01-01'
})
```


I'd love to be a contributor to such an awesome library. If you like the concept I could submit a PR?

akmjenkins:
[...]

PR #556, opened 2019-06-19
Allow access to parent schema (and unlimited ancestors!) in test context

### LINKED ISSUE #551: Access parent schema in test
It's easy enough to get the parent object in a custom test, but is there any way to get the parent schema? I have a use case where it would be useful to be able to `reach` into the parent schema. Just wondering if it's exposable, I tried to read through `src/mixed.js` but was unsuccessful at determining if it could be exposed.

akmjenkins:
I made a terrible attempt at this PR earlier. This should do it.

This resolves #551 and adds the capability to help with #553.

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -131,7 +131,7 @@ inherits(ObjectSchema, MixedSchema, {
     endEarly = this._option('abortEarly', opts);
     recursive = this._option('recursive', opts);
 
-    opts = { ...opts, __validating: true, originalValue };
+    opts = { ...opts, __validating: true, originalValue, parentSchema: this };
 
     return MixedSchema.prototype._validate
       .call(this, _value, opts)
@@ -156,6 +156,7 @@ inherits(ObjectSchema, MixedSchema, {
             ...opts,
             path,
             parent: value,
+            parentSchema: this,
             originalValue: originalValue[key],
           };
 
--- a/src/object.js
+++ b/src/object.js
@@ -128,11 +128,7 @@ inherits(ObjectSchema, MixedSchema, {
     let originalValue =
       opts.originalValue != null ? opts.originalValue : _value;
 
-    let fromClosure = (scope, from) => () => {
-      return { value: originalValue, schema: scope, from };
-    };
-
-    let from = fromClosure(this, opts.from);
+    let from = [{ schema: this, value: originalValue }, ...(opts.from || [])];
 
     endEarly = this._option('abortEarly', opts);
     recursive = this._option('recursive', opts);
@@ -149,6 +145,9 @@ inherits(ObjectSchema, MixedSchema, {
           return value;
         }
 
+        from = originalValue
+          ? from
+          : [...from].splice(0, 1, { schema: this, value: originalValue });
         originalValue = originalValue || value;
 
         let validations = this._nodes.map(key => {
--- a/src/object.js
+++ b/src/object.js
@@ -147,7 +147,10 @@ inherits(ObjectSchema, MixedSchema, {
 
         from = originalValue
           ? from
-          : [...from].splice(0, 1, { schema: this, value: originalValue });
+          : [
+              { schema: this, value: originalValue || value },
+              ...(opts.from || []),
+            ];
         originalValue = originalValue || value;
 
         let validations = this._nodes.map(key => {
--- a/src/object.js
+++ b/src/object.js
@@ -146,7 +146,7 @@ inherits(ObjectSchema, MixedSchema, {
         }
 
         from = originalValue
-          ? from
+          ? [...from]
           : [
               { schema: this, value: originalValue || value },
               ...(opts.from || []),
--- a/src/object.js
+++ b/src/object.js
@@ -128,10 +128,16 @@ inherits(ObjectSchema, MixedSchema, {
     let originalValue =
       opts.originalValue != null ? opts.originalValue : _value;
 
+    let fromClosure = (scope, from) => () => {
+      return { value: originalValue, schema: scope, from };
+    };
+
+    let from = fromClosure(this, opts.from);
+
     endEarly = this._option('abortEarly', opts);
     recursive = this._option('recursive', opts);
 
-    opts = { ...opts, __validating: true, originalValue, parentSchema: this };
+    opts = { ...opts, __validating: true, originalValue, from };
 
     return MixedSchema.prototype._validate
       .call(this, _value, opts)
@@ -155,8 +161,8 @@ inherits(ObjectSchema, MixedSchema, {
           let innerOptions = {
             ...opts,
             path,
+            from,
             parent: value,
-            parentSchema: this,
             originalValue: originalValue[key],
           };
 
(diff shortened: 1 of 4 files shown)


src/object.js
```diff
@@ -128,10 +128,16 @@ inherits(ObjectSchema, MixedSchema, {
     let originalValue =
       opts.originalValue != null ? opts.originalValue : _value;
 
+    let fromClosure = (scope, from) => () => {
+      return { value: originalValue, schema: scope, from };
+    };
```
jquense: this might be a bit simpler if instead of a nested closure, ancestors were stored in an array? e.g. `this.ancestors[0]`

akmjenkins: > this might be a bit simpler if instead of a nested closure, ancestors were stored in an array? e.g. `this.ancestors[0]`

I thought about that as well as an option, but I felt an API of
```
   const parent = from[0];
   const grandparent = from[1].value;
```

to feel less intuitive of:
```
   const parent = from();
   const grandparent = from().from();
```

But that's very much personal preference. It's your library, and if you like the concept, I'd be happy to update the PR!

Are you still okay with accessing the parentSchemas so it would look like:

```
const { value, schema } = from[0];
const { value: grandparentValue, schema: grandparentSchema } = from[1];
```

BenevidesLecontes:
@jquense we need this desperately.

src/object.js
```diff
@@ -143,6 +145,9 @@ inherits(ObjectSchema, MixedSchema, {
           return value;
         }
 
+        from = originalValue
+          ? from
+          : [...from].splice(0, 1, { schema: this, value: originalValue });
```
jquense: `splice` returns the removed item not the array, is that what you want here?

src/object.js
```diff
@@ -143,6 +145,9 @@ inherits(ObjectSchema, MixedSchema, {
           return value;
         }
 
+        from = originalValue
+          ? from
```
jquense: should probably not overload the variable here so children get a fresh, new array, incase the mutate it it won't break anything down the line?

jquense:
what happens if you have an array inbetween?

src/object.js
akmjenkins: No, it's not what I wanted to do, thanks for the catch. I've updated this.

src/object.js
akmjenkins: Would this be correct?

```
from = originalValue ? [...from] : ...
```

jquense: ya

test/mixed.js
```diff
@@ -483,6 +483,64 @@ describe('Mixed Types ', () => {
...
+    let third = object({
+      thirdField: mixed().test({
+        test() {
+          calledThird = true;
+
+          // I want to do something like this,
+          // this.resolve(ref(this.path),this.from[1].value);
+          // OR probably preferably
+          // this.from[1].resolve(ref(this.path))
+          // but I can't get it working :( ...help?
```
jquense: what are you trying to test with that broadly? I'd think you want handle ref resolving differently on top of `from` existing. e.g. `ref('../../firstField')`

akmjenkins: I could only determine if the `from[0].value` is correct when dealing with an array, if 

1. I could resolve the ref or
2. I added the index option [here](https://github.com/jquense/yup/pull/556/files/fa036aa8981567d37ac647bd71b8e5a7573f85e5#diff-dfa1d8cd44cb64e7aff89d961845327dR92)

I didn't particularly like the index option, so I was trying to resolve the ref. If you don't mind leaving that index option in the test context when dealing with an array, then you're right, we can figure out how to resolve the relative ref later.

akmjenkins: You know what you are when dealing with a parent object, because you know you're own key, but you don't know what you are when dealing with a parent array - hence the idx. Take a look at [this line](https://github.com/jquense/yup/pull/556/files/fa036aa8981567d37ac647bd71b8e5a7573f85e5#diff-a39772825a6753e21b910a44bde28402R514).

Sorry for being AWOL for a while, I'm still a big fan of adding this feature to yup, just want to make sure it's done right.

yanzou:
@akmjenkins any update on this? 
it takes me 3+ misrable days googleing/check issues/the Yup source code and realize it is not possible to get values from ancestors...

Any reason we can't do this in the first place?  Isn't that a universal requirement for validaion?

test/mixed.js
akmjenkins: @jquense if you are fine with keeping the index option, then I can remove this comment from the test code and resolve this conversation?

jquense: yup seems ok I think

test/mixed.js
```diff
@@ -483,6 +483,64 @@ describe('Mixed Types ', () => {
...
+          this.from[0].schema.should.equal(first);
+          return true;
+        },
+      }),
+      second,
+    });
+
+    await first.validate(testFixture);
+    calledFirst.should.equal(true);
+    calledThird.should.equal(true);
```
jquense, line 541: I'm still find this test hard to understand. Maybe b/c it's large. If the goal is to assert that the `from` value is correct maybe instead of all the flags, have the deeply nested schema save off the `from` value via a closure then assert after the validate here that it matches what you expect?

akmjenkins: Done!

HansAarneLiblik:
This could really use documentation

conioX:
i need documenation for this

geocine:
For anyone wondering how this could be used this is how.
```js
  const schema = Yup.object().shape({
    origin: Yup.object().shape({
      location: Yup.string().required(),
      port: Yup.string()
        .nullable()
        .test('customValidation', 'Error', function(val) {
          const { parent, from } = this;
          
          // Before we can only access the immediate parent values
          // of this property which is `origin` so we can get `parent.location`

          // But what if we want to get `destintation.location` , parent is not possible
          // now we can using, `from[1].value.location` for getting the value 
        })
    }),
    destination: Yup.object().shape({
      location: Yup.string().required(),
      port: Yup.string().required()
    })
  });
```
So we can zoom from the context of `origin.port` above. If you need `n` levels up just use `from[n]` where `n` is the number of levels starting from `0`

So getting back at the example above `parent.location` will be the same as `from[0].value.location`

There was also no way of getting the `schema` now you can get the `schema` at different levels using `from[n].schema`.

zelucena:
I have just come across this issue. I upgraded Yup in my project from version 0.28.1 to 0.29.1
`const { parent, from } = this;
        const grandParent = from[1].value;`
It works just fine.

lucksp:
Are the typescript Types updated?  I do not see anything with `from` declared, so I am having to do some nasty casting...

adaladam:
awesome work guys, thank you!

singlapradeep27:
will it work for when too?

adjenks:
> will it work for when too?

You can try stuffing the values object into the context parameter during validation, then accessing the values with the "$" context prefix. This worked for me, but it seems to have some limitations of its own.

PR #561, opened 2019-06-25
docs: Add usage from TypeScript

mauricedb:
Show how to use Yup with TypeScript and how to use the `InferType` to create a type alias from a Yup schema.


jquense:
thanks!

PR #576, opened 2019-07-14
Example of custom type defuntion with mixed()

mauricedb:
Added an example of using `yup.mixed<T>()` to use custom type definitions with TypeScript.


jquense:
thanks!

PR #579, opened 2019-07-23
feat: provide keys in default noUnknown message

doomsower:
Currently `noUnknown` does not provide the list of unknown key, which makes it very hard to debug. This PR fixes it.

### DIFF
--- a/src/locale.js
+++ b/src/locale.js
@@ -53,7 +53,7 @@ export let date = {
 export let boolean = {};
 
 export let object = {
-  noUnknown: '${path} field cannot have keys not specified in the object shape',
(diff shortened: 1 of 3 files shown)


andrey-hohlov:
🙏

jquense:
missed this! great addition thanks!

msheakoski:
Thanks for this PR! I ran into this issue today. @jquense would it be possible to publish a release with this change?

I attempted to use `https://github.com/jquense/yup#ad5d015` as the version in my package.json but did not have any luck since the package needs to be built to generate a lib/index.js

PR #598, opened 2019-08-11
fix: synchronous conditional object validation with unknown dependencies

mikaelkolkinn:
This PR fixes an issue where using `when` on an object with dependencies on properties that were not specified in the schema, caused the validation to become asynchronous even with `sync` set to true.


jquense:
Awesome thanks!

PR #607, opened 2019-08-20
.test methods now accept translation for error messages

Bossuletz:


jquense:
hey there, thanks for the PR. Not sure i understand what the change is doing tho the title suggests its for translation. How do having messages in `locale.test` help

Bossuletz:
Hi there,
I have a complex validation schema (which includes using `test` validation). Also, my app is translated in many languages, so I need that all the errors to be translatable. For other functions, like `string` or `required`, the translation is already in place, but `test` doesn't. I've added a way to also add translations to test, by using the test name as a unique key.

jquense:
Gotcha, I don't think this makes sense as a general feature since it's fairly specific to your implementation and structure. For custom validations I suggest taking the same approach as all the built in validations which is to set the default message for any specific validation in the method itself.

PR #633, opened 2019-09-13
Fix insertion order when key matches partially

lourenci:
When a key of an object matches partially with another key, the lib is not respecting the defined order.

I created a new spec context to clarify and to prevent someone to misunderstand the `barfoo` key and delete it. Pls, let me know if you don't see a need for that.

### DIFF
--- a/src/util/sortByKeyOrder.js
+++ b/src/util/sortByKeyOrder.js
@@ -1,7 +1,7 @@
 function findIndex(arr, err) {
   let idx = Infinity;
   arr.some((key, ii) => {
-    if (err.path.indexOf(key) !== -1) {
+    if (new RegExp(`^${err.path}$`).test(key)) {
       idx = ii;
       return true;
     }
--- a/test/object.js
+++ b/test/object.js
@@ -586,11 +586,12 @@ describe('Object types', () => {
     ]);
   });
 
-  it('should sort errors by insertion order', async () => {
+  fit('should sort errors by insertion order', async () => {
     let inst = object({
       // use `when` to make sure it is validated second
       foo: string().when('bar', () => string().min(5)),
       bar: string().required(),
+      barfoo: string().required(),
     });
 
     let err = await inst
@@ -600,6 +601,7 @@ describe('Object types', () => {
     err.errors.should.eql([
       'foo must be at least 5 characters',
       'bar is a required field',
+      'barfoo is a required field',
     ]);
   });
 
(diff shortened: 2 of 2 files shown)


src/util/sortByKeyOrder.js
```diff
@@ -1,7 +1,7 @@
 function findIndex(arr, err) {
   let idx = Infinity;
   arr.some((key, ii) => {
-    if (err.path.indexOf(key) !== -1) {
+    if (new RegExp(`^${err.path}$`).test(key)) {
```
jquense: won't this break for complex paths? like: `'foo.bar['baz']'`

lourenci: All the tests have passed. How can I test this? My yup usage is very basic with formik.

jquense: I'd adjust your test to use a nested object. 

```
object({
  foo: object({
    bar: string()
  })
})
```

lourenci: Done. Take a look at it, pls.

lourenci: Hi, @jquense. Do you have any other consideration about this?

lourenci: Sorry for the ping, @jquense, but is there something I can do to get this merged?

lourenci:
Hi, @jquense. Do you have any plan to merge this into master?

src/util/sortByKeyOrder.js
jquense: sorry for forgetting this. I don't think this is right still `new RegExp(`^${err.path}$`)` will create a regular expression that matches slightly wrong things, because paths contain regex tokens that would need to be escaped, in particular `.`. However, the regex here isn't doing anything other than `a === err.path`, which is to say it matches the entire string exactly (or would if the path was escaped). 

This also won't work if the nested object needs it's keys sorted:

```js
object({
  nested: object({
    foo: string().when('bar', () => string().min(5)),
    bar: object({
      foo: string().required(),
    }),
    barfoo: string().required(),
  }),
});
```

When findIndex runs on `nested`'s children the field array will be `['foo', 'bar', 'barfoo'] but the `error.path` will start with `nested` e.g. it will be 'nested.foo` which won't match any paths. Honestly i'm not sure how to solve this, but it seems like i'd be more complicated than is worth it, but definitely open to ideas!

PR #635, opened 2019-09-19
Multiple dynamic languages support for yup

cipacda:
I've added support for yup to be used with multiple language locales set. This is very helpful in an environment with multiple languages supported at the same time, i.e. node.js express application.

Can you please have a look at this and make a new release of it, given you're happy with it? I've updated all tests, tried to be backwards compatible.

This is blocking issue for us at the moment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,8 +110,8 @@
   "dependencies": {
     "@babel/runtime": "^7.10.5",
     "@types/lodash": "^4.14.165",
-    "lodash": "^4.17.20",
-    "lodash-es": "^4.17.15",
+    "lodash": "^4.17.21",
+    "lodash-es": "^4.17.21",
     "nanoclone": "^0.2.1",
     "property-expr": "^2.0.4",
     "toposort": "^2.0.2"
--- a/yarn.lock
+++ b/yarn.lock
@@ -2934,9 +2934,9 @@ camelcase@^6.0.0:
   integrity sha512-8KMDF1Vz2gzOq54ONPJS65IvTUaB1cHJ2DMM7MbPmLZljDH1qpzzLsWdiN9pHh6qvkRVDTi/07+eNGch/oLU4w==
 
 caniuse-lite@^1.0.30001035, caniuse-lite@^1.0.30001093:
-  version "1.0.30001119"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001119.tgz#99185d04bc00e76a86c9ff731dc5ec8e53aefca1"
-  integrity sha512-Hpwa4obv7EGP+TjkCh/wVvbtNJewxmtg4yVJBLFnxo35vbPapBr138bUWENkb5j5L9JZJ9RXLn4OrXRG/cecPQ==
+  version "1.0.30001208"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001208.tgz"
+  integrity sha512-OE5UE4+nBOro8Dyvv0lfx+SRtfVIOM9uhKqFmJeUbGriqhhStgp1A0OyBpgy3OUF8AhYCT+PVwPC1gMl2ZcQMA==
 
 capture-exit@^2.0.0:
   version "2.0.0"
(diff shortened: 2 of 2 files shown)


jquense:
Thanks for taking the time to work on this. I'm a bit reticent to really codify locales as part of yup, there is a lot of complexity around it that can get out of hand and a lot of nuance that is out of scope (like choosing the best fit culture, etc).

What if instead of setting up multiple languages, you could just pass in the messages at the `validate()` call instead? That would simplify the code here, and push more complexity back to the user app. I will also add that we use yup in multi language environments without any of this, by setting messages to the message key, instead of the string and mapping back to the translated string when we want to display something.

cipacda:
Passing all messages at the validate call looks a bit like a code smell, adding a lot of redundancy in the user app. Having a single point where we can configure yup is a very good addition, and we can setup only once per application.  With your suggestion, in a multi-language express backend app, we'd need to pass the request context to the middleware doing the schema validation, so we know what language to use.

Similarly, we cannot pass message by passing them to the message key, as we'd need to define the schemas at each request (i.e. by having a callback or something), which will be a performance problem sooner or later. 

I think that if we have all the locales set up by yup, and we tell yup only the locale to use at the validate point, we simplify the user experience, without having to add a lot of logic to yup - we can be clear that yup does not handle any locales complexity - like choosing the best fir culture, etc.

I've been using this successfully on our medium size, but quickly growing project, and it helped us a lot in only one week.  I am yet to find downsides to this approach, and it is still fully backwards compatible for people not needing.

However, I am happy to other suggestions on how to work around this, if there's something I don't see, without needing to have a wrapper on top of yup.

jquense:
> Similarly, we cannot pass message by passing them to the message key, as we'd need to define the schemas at each request (i.e. by having a callback or something), which will be a performance problem sooner or later.

Sorry not sure i see how this would happen...we do something like:

```
setLocale({
  string: {
    required: { id: 'validation.required }
  }
})
```

and then something like `translate(currentCulture, validationErrror)` when passed out or rendered. The key thing here is that yup has no idea what the strings are and doesn't need to care. I'm not sure what your i18n infrastructure looks like, but this seems to jive with most patterns, where strings are given a stable ID, that is mapped back to a locale hash for the current language.

Overall the real concern is that _most_ people use yup in UI's on the client and so we need to keep the library lean, and not codify to many specific patterns to accommodate that use case. I do want the server-side case to work great, so glad to have your experience and insight to talk through this. I think ideally we come up with something that doesn't require a secific pattern of localization, while also meeting the needs of doing i18n

cipacda:
> Sorry not sure i see how this would happen...we do something like:

I misunderstood what you were suggesting first, thought that you were suggesting to pass actual translated messages.

Our i18n infrastructure is similar to what you're suggesting and we're using .json files to map from stable IDs to the locale. 

I don't necessarily like the idea to go through every message parsed already by yup (i.e. taken from the yup.json file) and then translate it. The above makes it impossible to pass translation files directly to yup, for it to handle - which is functionality currently supported, but for only one locale!

My suggestion would work similarly well for both UI and backend - the UI users should do nothing differently to what they're doing now - we are actually using yup on both the UI and the backend, and my current PR works well.

Passing a key as a message in the locale setup, would make it also complex for the user app to translate interpolated strings - i.e. `${path} must be one of the following values: ${values}'`

I am happy if you have another suggestion on how to implement this (to the same outcome), and am also happy to amend my PR accordingly, as I do need this functionality, but I thought this mimics mostly how i18n works as well and it provides an easy integration with i18next whilst keeping it clean and similar for all current users and I am not sure it provides any extra complexity - I even moved all the message composition in one place, before it's passed back to the user app.

tafelito:
@jquense I know this is an old thread, and even tho I found that setting a message key when setting up the locale could work for some cases, I couldn't find a good way of also translating the path part of the message. 

Most of the times, if not all the time, we need to not only translate the message but also the path where the error originated. 
Having a message like `${path} must be one of the following values: ${values}'` the path could be something like `firstName`. We don't want to show the camelCase format of the path, so we could use instead a function like suggested in the docs and pass the path as an argument and then translate that as well. 
The problem with that approach is when you have a nested path, for example in an array. So the path then becomes something like this `users.[2].firstName`. 
Now we would have to add some logic to parse those special cases. 
I wonder, since you mentioned that you have worked with multi language environments how you handle those cases. 
For some custom methods that we added, we can add some data to the meta and use that instead of the path, but when using function in setLocale, you don't have access to the meta property there.

Issue #153, opened 2017-12-06
Error when using required and nullable


ScreamZ:
Hey,

I have an object that can is by default undefined, and accept one of `null` or a `string`value.

I achieve the desired behavior using

````js

Yup.string()
  .nullable()
  .oneOf([null, 'female', 'male'])
  .notOneOf([undefined])
````

that should be the same than:

```js
Yup.string()
  .nullable()
  .required()
  .oneOf([null, 'female', 'male', 'other'])
```

But this is not...

if I read the manual `required` is checking for undefined and `nullable` for null, but they seems to conflicts each other...

Can you check that ?

best regards

jquense:
required actually checks "presence" which is `undefined` or `null` since that generally what folks mean by required. We should probably add a `defined()` check as well tho for the case where you just don't want it to be `undefined`. in the meantime it's easy to add

```js
yup.addMethod(yup.mixed, 'defined', function () {
   return this.test('defined', "{path} must be defined", value => value !== undefined)
})
```

jquense:
> do you want me to make a PR for both those features ?

that would be wonderful :)

DesignByOnyx:
Anybody wishing to use the "defined" method along with TypeScript should also add this:

```ts
declare module 'yup' {
	export interface Schema<T> {
		defined<U>(): this;
	}
}
```

Issue #189, opened 2018-03-20
Why are empty arrays considered missing values?


DylanVann:
An empty array doesn't mean a value is missing. It means the value is an empty array.

So I'm validating responses coming back from a server.

Valid:
- empty array
- array of strings

Not valid:
- undefined

I want to throw an error if the field is missing because it means something is wrong and the client and server are out of sync. Maybe they changed the property name on the server.

jquense:
> An empty array doesn't mean a value is missing

No one is claiming that it does, the tests is "required" not "present". The current behavior better matches what folks expect when you say "this list of things is required", same as with strings. It's not testing whether it's defined.  

You can get the behavior your looking for with a custom test, check out for an example: https://github.com/jquense/yup/issues/153#issuecomment-349679982

DylanVann:
In the readme:

> except that empty arrays are also considered 'missing' values.

In my test I'm checking that it's there, and that it's an array, because I want to call things like `.map` on it later. If the value is `[]` I can safely call map on it.

If I call something like `GET /articles` with some parameters I'd expect it to return an array, but also that the array could be empty.

It seems like odd behavior to me. Thank you so much for the quick response and pointing out a workaround though.

jquense:
hmm poor wording in the docs then, thanks for pointing it out. I'm disagreeing that the behavior you want is useful, but evidence i've seen says the way we have it is more intuitive for common use cases. 
I'm happy to have both behaviors in the linked Issue i note that I'd definitely accept a PR adding a "defined" method

tms1337:
For me as well the array is present even if it's empty
For example if I ask for a list of items in a cart, even if empty it is still valid list of items.
Only if it's `null` or `undefined` does it make issues.

rsp:
@DylanVann This is indeed a very confusing behavior.

The current behavior of `array().required()` is what I would expect from `array().required().minLength(1)` or `array().required().nonEmpty()` and because of that I almost never use `array().required()` because in the real world most arrays can be empty but we want to be sure that they are valid arrays to that things like `.map()` and `.filter()` work. So in practice I end up using things like `array().test(value => Array.isArray(value));` which is unfortunate that such a simple thing is not supported out of the box.

jquense:
You can use `array().defined()`

tms1337:
@jquense is that equivalent to `array().test(value => Array.isArray(value));` ?

douglas-treadwell:
> You can use `array().defined()`

[...]

PR #637, opened 2019-09-25
feat: Add defined() validation to mix

talyssonoc:
After seeing this comment https://github.com/jquense/yup/issues/153#issuecomment-349679982 in an issue about the error when mixing `required` with `nullable` I thought it would be nice to have this validation.

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -540,6 +540,17 @@ const proto = (SchemaType.prototype = {
         ),
     };
   },
+
+  defined(message = locale.defined) {
+    return this.nullable().test({
+      message,
+      name: 'defined',
+      exclusive: true,
+      test(value) {
+        return value !== undefined;
+      },
+    });
+  },
 });
 
 for (const method of ['validate', 'validateSync'])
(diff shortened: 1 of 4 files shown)


src/mixed.js
```diff
@@ -540,6 +540,21 @@ const proto = (SchemaType.prototype = {
...
+    return value !== undefined;
+  },
+
+  defined(message = locale.defined) {
+    return this.nullable().test({
+      message,
+      name: 'defined',
+      exclusive: true,
+      test(value) {
+        return this.schema._isDefined(value);
```
jquense: I don't think we need the indirection here, since it's unlikely anyone will override this. Let's just do the check here?

talyssonoc: Sure, gonna do it!

Also, do you think I should implement a version of it for arrays as well? I ask it based in this comment https://github.com/jquense/yup/issues/189#issuecomment-374701160

jquense:
thanks!

src/mixed.js
dwjohnston: @jquense  @talyssonoc  - Yes please for arrays - I exactly need this.

dwjohnston: ^Wait - does this need more work for arrays? Seems to be working fine for me.

Issue #644, opened 2019-10-03
Documentation for setLocale doesn't mention function support


fmoliveira:
**Describe the bug**
The documentation for the [Using a custom locale dictionary](https://github.com/fmoliveira/yup#using-a-custom-locale-dictionary) doesn't mention that it's possible to use functions to generate error messages or error objects.

**To Reproduce**
Check the documentation for [Using a custom locale dictionary](https://github.com/fmoliveira/yup#using-a-custom-locale-dictionary) and try to find where it mentions that `setLocale` accepts defining customised messages with functions.

**Expected behavior**
The documentation for [Using a custom locale dictionary](https://github.com/fmoliveira/yup#using-a-custom-locale-dictionary) should include an example of how to customise the messages using functions to make the output optimal for i18n libraries.

**Additional context**
This feature is especially useful when using Yup in a project with i18n, where we will need Yup to return a translation key instead of the final message, along with the values to be attached to the translations. We have this logged in an issue at https://github.com/jquense/yup/issues/159#issuecomment-453737260 , but this should ideally be documented in the readme.

fmoliveira:
PR submitted at https://github.com/jquense/yup/pull/645

PR #645, opened 2019-10-03
Add setLocale example with functions

fmoliveira:
Fixes https://github.com/jquense/yup/issues/644

This will include in the readme an example of how to customize messages using functions, which is especially useful for projects with i18n.


jquense:
awesome thanks!

PR #655, opened 2019-10-15
fix: unix epoc bug in date parser

### LINKED ISSUE #234: Date 1970-01-01T00:00:00Z is not validating.
It looks like a valid date, and standard JS Date object can parse it.

```js
new Date('1970-01-01T00:00:00Z')
```

However, `yup` validator fails:

https://runkit.com/moltar/5b293e490d95e300127736f4

> TypeError: The value of createdOn could not be cast to a value that satisfies the schema type: "date". 
> 
> attempted value: "1970-01-01T00:00:00Z" 
> result of cast: null

Thanks!

privateOmega:
Hopefully this fixes #234 and doesn't introduce any new buggy conditions.

I have modified return `value ? new Date(value) : invalidDate;` to `return value > -1 ? new Date(value) : invalidDate;` keeping in mind any valid values will be greater than -1, (i.e 0 or above) and the invalid value like `NaN` doesn't satisfy `> -1` condition.

### DIFF
--- a/src/date.js
+++ b/src/date.js
@@ -21,7 +21,8 @@ function DateSchema() {
       if (this.isType(value)) return value;
 
       value = isoParse(value);
-      return value ? new Date(value) : invalidDate;
+      // 0 is a valid timestamp equivalent to 1970-01-01T00:00:00Z(unix epoch)
+      return value > -1 ? new Date(value) : invalidDate;
     });
   });
 }


jquense:
awesome thanks!

brandon-leapyear:
https://github.com/jquense/yup/issues/731

PR #669, opened 2019-10-22
Fix array path resolve for descendants

serhii-ohorodnyk:
At the moment during casting process Array type passes the field path to sub items without resolving it according to sub item array index.
The path is exposed to end consumer as part of context in conditional validation `when` (could be somewhere else).
If conditional validation relies on context path - this behaviour will lead to bugs.

### DIFF
--- a/src/array.js
+++ b/src/array.js
@@ -47,12 +47,10 @@ inherits(ArraySchema, MixedSchema, {
 
     let isChanged = false;
     const castArray = value.map((v, idx) => {
-      var path = makePath`${_opts.path}[${idx}]`;
-      var innerOptions = {
+      const castElement = this._subType.cast(v, {
         ..._opts,
-        path,
-      };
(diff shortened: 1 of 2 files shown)


src/array.js
```diff
@@ -46,8 +46,13 @@ inherits(ArraySchema, MixedSchema, {
     if (!this._typeCheck(value) || !this._subType) return value;
 
     let isChanged = false;
-    const castArray = value.map(v => {
-      const castElement = this._subType.cast(v, _opts);
+    const castArray = value.map((v, idx) => {
+      var path = makePath`${_opts.path}[${idx}]`;
+      var innerOptions = {
```
jquense: can you move this inline below: `cast(v, { ..._opts, path: makePat`

serhii-ohorodnyk: Thanks for quick check @jquense ! Made the change.

jquense:
thanks!

Issue #71, opened 2016-11-28
Pluggable i18n locale/translate


rosskevin:
I see the locale messages in https://github.com/jquense/yup/blob/master/src/locale.js

But there doesn't appear to be any way to alter these messages, nor a way to plug-in other i18n solutions by referring simply to keys and allowing it to resolve independently.

Is this something others are needing too?

rosskevin:
After a cursory review, it would seem to need the following to make this happen:
1. Update [`mixed.validate`](https://github.com/jquense/yup/blob/master/src/util/createValidation.js) to accept a `translate(key)` function
2. Update all schema code to pass the locale key instead of a resolved string message for errors
3. Update [`createValidation`](https://github.com/jquense/yup/blob/master/src/util/createValidation.js#L25) to resolve the message using the given translate or fallback to the hardcoded locale

Does this look correct?  Are you open to a PR on this?

jquense:
hey there, sorry I don't have a moment right now to really comment, but there is two things here. "yes, extending/altering the existing language is too hard", and "no we don't necessarily need to add a `translate` sort of functionality". When I get a moment I'll post more about how we handle i18n with yup/react-formal

rosskevin:
When you get a moment, just so I'm clear, I'm not saying implement translate within yup, but allow a `translate(key)` function to be provided by the user.  

I'm open to anything, we use i18next for our app.

jquense:
So `yup` doesn't care what shape your messages are. you can just provide a key if you want.

by default react-formal _expects_ the message to a string, but that's not required. We use `react-intl` which has message objects that like like `{ id: 'someKey', defaultMessage: 'foo' }`. We override all the provided messages with these sort of objects as well as for any other messages.

react-formal `Messages` have an `extract` which maps the yup error json to a string to display, and defaults to: `error => error.message || error,`. You can change that to handle translateing the key to a real message like:

```js
export default class FormMessage extends React.Component {
  static contextTypes = {
    intl: intlShape.isRequired,
  };

  extract = (error) => {
    const key = error.message || error
    return this.context.intl.formatMessage(key);
  };

  render() {
    return (
      <Form.Message
        {...this.props}
        extract={this.extract}
      />
    );
  }
}
```

jquense:
what _is_ hard about yup right now tho, is that in order to override the default locales you need to require and mutate the yup one, which is gross. I'd happily accept a PR to give that a proper API.

rosskevin:
[...]

PR #677, opened 2019-10-26
Look up default error messages for custom validations in Locale

the-daniel-rothig:
When a custom validation is defined by invoking `mixed().test('myCustomValidation', () => {...})`, yup checks the locale for a field matching the test name.

This allows us to write libraries of new validations while keeping control of messages in the client code.

For example, I could in a library

```
// ./node_modules/yup-extensions/index.js
import { addMethod, string } from yup;

addMethod(string, 'mustNotContain', function(stringToExclude, message) {
  return this.test({
    name: 'mustNotContain',
    test: (value) => value.indexOf(stringToExclude) === -1,
    message: message,
    params: {
      mustNotContain: stringToExclude
    }
  })
})
```

And this library could be used by a client as follows:

```
// clientCode.js

import { setLocale, string } from yup;
import 'yup-extensions';

setLocale({
  string: {
    required: "Please enter a value for ${path}",
    mustNotContain: "${path} must not contain '${mustNotContain}'"
  }
})

const validator = string()
  .required()
  .mustNotContain("<")
```

This way, the client need not worry if the method is part of the `yup` core library or a custom method: their interface for locale setting is identical.

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -1,8 +1,7 @@
 import has from 'lodash/has';
 import cloneDeepWith from 'lodash/cloneDeepWith';
 import toArray from 'lodash/toArray';
-
-import { mixed as locale } from './locale';
+import globalLocale, { mixed as locale } from './locale';
 import Condition from './Condition';
 import runValidations from './util/runValidations';
 import prependDeep from './util/prependDeep';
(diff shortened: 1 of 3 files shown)


nealoke:
@jquense can we please get this merged, it seems so obvious to include 😄

reecelucas:
This looks really useful. @jquense it would be great to get this in at some point.

the-daniel-rothig:
For anyone interested in getting this functionality today, I have written a small package that does exactly what this PR does. It also works with existing extension libraries.

https://www.npmjs.com/package/yup-universal-locale

But hopefully this package will be redundant soon ;)

cinnabarcaracal:
The `required` part of the above example is currently not supported:

```
setLocale({
  string: {
    required: "Please enter a value for ${path}", // <-- this part
    mustNotContain: "${path} must not contain '${mustNotContain}'"
  }
})
```

https://github.com/jquense/yup/issues/71

zeing:
I wanna this, easy for setup error msg from 3rd party i18n lib

PR #703, opened 2019-11-23
Added multipleOf test for numbers

tecfu:
Adds feature to increase compatibility with OpenApi3 spec.


src/number.js
```diff
@@ -91,6 +91,18 @@ inherits(NumberSchema, MixedSchema, {
...
   },
 
+  multipleOf(multipleOf, message = locale.multipleOf) {
+    return this.test({
+      message,
+      name: 'multipleOf',
+      exclusive: true,
+      params: { multipleOf },
+      test(value) {
+        return isAbsent(value) || value % this.resolve(multipleOf) === 0;
```
msairio: I had issues with multipleOf less than 1. With value 1.1 and multipleOf 0.1 the test fails as `1.1 % 0.1 === 0` is false. This is apparently do to javascript floating point precision errors. 

The problem could be worked around in a hacky manner by multiplying both sides of modulo test to integers using multiplier of `Math.pow(10, -Math.floor(Math.log10(multipleOf)))`. Resulting test `11 % 1 === 0` becomes true.

tecfu: You could simply solve this by multiplying the dividend and the divisor by the maximum number of decimal places you wish to support. I chose to go with the maximum available to javascript (6, before hitting scientific notation and losing precision) in [@tecfu/is-multiple-of](https://github.com/tecfu/is-multiple-of), a package I published to solve the floating point issue. 

I pushed a new commit (bc7608e) using that package, along with additional unit tests.

If you don't want to use it but still employ this feature for Swagger compliance, you'll also want to throw an error when large or small numbers are passed in scientific notation, as they will always return a false positive. 

Lastly, you'll need to account for 0 being a multiple of only 0.

Issue #708, opened 2019-11-27
`this.createError()` accepts a `params` key, but the docs and TS types say otherwise


ericbiewener:
The docs and typings suggest that you can only provide `path` and `message` keys in the argument passed to `this.createError(...)`. However, I'm able to provide `params` as an argument and have those picked up by the interpolated message. [Looking at the source code](https://github.com/jquense/yup/blob/master/src/util/createValidation.js#L41) this appears to be supported, so my guess is we just need an update to the docs & typings. Happy to open a PR if this is indeed correct.

jquense:
I think it's publicly supported, happy to take a PR adding to the docs!

PR #709, opened 2019-11-27
Document optional params property in createError

ericbiewener:
Addresses #708. Will open a PR in DefinitelyTyped to fix the TS definition of `createError`.


jquense:
thanks!

PR #710, opened 2019-11-28
feat: add runInOrder function so the validations can run in order

### LINKED ISSUE #702: Add an option so validations can run in sequence
With a form field, I'd like the former validation function can terminate the whole chain when it already throws error, so the later function especially with ajax calls won't be called.

For example, add a `sequential` option which can be set to true, and I changed some code in `src/mixed.js` to explain this more clearly.
```js
let endEarly = this._option('abortEarly', options);
...
// get the sequential option
let sequential = this._option('sequential', options)
```
```js
if (this._blacklistError)
      initialTests.push(this._blacklistError(validationParams));
...
// get validations
let validations
if (endEarly && sequential) {  // this may need to work with abortEarly
    validations = [new Promise((resolve, reject) => {
        let p = this.tests[0](validationParams)
        for (let i = 1; i < this.tests.length; i++) {
            p = p.then(() => this.tests[i](validationParams))
        }
        p.then(resolve).catch(reject)
    })]
} else {
    validations = this.tests.map(fn => fn(validationParams))
}
```
```js
...
validations: this.tests.map(fn => fn(validationParams)), \* changes to *\ validations,
...
```

ongw20:
example:
```js
const schema = string()
      .required()
      .test('a', 'a', (val) => {
        return false
      })
      .test('b', 'b', (val) => {
        return false
      })
      .runInOrder()
      .test('c', 'c', (val) => {
        return false
      })
```
`required` `a` and `b` are run in sequential, while c won't be affected.
when you run
```js
schema.validate('name', {abortEarly: false})
      .then(val => console.log(val))
      .catch(err => console.log(err))
```
and the output is
```js
ValidationError {
      name: 'ValidationError',
      value: 'name',
      path: undefined,
      type: undefined,
      errors: [ 'a', 'c' ],
      inner: [
        ValidationError {
          name: 'ValidationError',
          value: 'name',
          path: undefined,
          type: 'a',
          errors: [Array],
          inner: [],
          message: 'a',
          params: [Object]
        },
        ValidationError {
          name: 'ValidationError',
          value: 'name',
          path: undefined,
          type: 'c',
          errors: [Array],
          inner: [],
          message: 'c',
          params: [Object]
        }
      ],
      message: '2 errors occurred'
    }
```

close #702

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -522,6 +522,38 @@ const proto = (SchemaType.prototype = {
     return next;
   },
 
+  runInOrder() {
+    const next = this.clone();
+    let tests = next.tests;
+    const fn = params => {
+      const schema = params.schema;
+      tests = tests.filter(fn => {
+        const test = schema.tests.find(
+          test => fn.OPTIONS.name === test.OPTIONS.name,
+        );
+        if (test) {
+          if (
+            test.OPTIONS.isExclusive ||
+            schema._exclusive[fn.OPTIONS.name] ||
+            fn.OPTIONS.test === test.OPTIONS.test
+          ) {
+            return false;
+          }
+        }
+        return true;
+      });
+      return tests.reduce((prev, next) => {
+        const _prev = typeof prev === 'function' ? prev(params) : prev;
+        return _prev.then(() => next(params));
+      });
+    };
+    fn.OPTIONS = {
+      name: () => 'run-in-order', // function to make name unique
+    };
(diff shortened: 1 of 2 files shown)


jquense:
Sorry I didn't get to this :/ GitHub is really bad at resurfacing PRs once you dismiss the notifications once. Thanks for taking the time to work on this even if we didn't get it in!

PR #715, opened 2019-12-04
feat: codepointLength, codepointMin, codepointMax

hlian:
## STRINGS

Hi! I was using Yup on a project (love the library), and I noticed it was counting UTF-16 bytes (`str.length`), which made it a bad fit for my project.

While there is [no](https://blog.jonnew.com/posts/poo-dot-length-equals-two) single [great](https://hsivonen.fi/string-length/) way to count string length, I think it would at least make sense to provide a _codepoint_-aware version of `length`, `min`, and `max`. It's 

- Short _`Array.from(str).length`_
- And fairly accurate _It doesn't account for Unicode's astral plane – but it correctly handles Unicode codepoints that can't be encoded into a single UTF-16 byte, which is the major issue with the non-codepoint ones_

Love the library. Hope y'all will consider landing this or some version of this!


src/string.js
```diff
@@ -137,8 +175,8 @@ inherits(StringSchema, MixedSchema, {
   },
 
   uppercase(message = locale.uppercase) {
-    return this.transform(
-      value => (!isAbsent(value) ? value.toUpperCase() : value),
+    return this.transform(value =>
+      !isAbsent(value) ? value.toUpperCase() : value,
```
hlian: The husky pre-commit hook caught some style errors, so I thought I'd land them in this commit too. I can also disable the precommit hook and retry if this is too noisy

src/locale.js
```diff
@@ -26,6 +26,9 @@ export let string = {
   length: '${path} must be exactly ${length} characters',
   min: '${path} must be at least ${min} characters',
   max: '${path} must be at most ${max} characters',
+  codepointLength: '${path} must be exactly ${length} characters',
+  codepointMin: '${path} must be at least ${min} characters',
+  codepointMax: '${path} must be at most ${max} characters',
```
hlian: i thought it'd be best to not mention codepoints to end users heh

jquense:
hey thanks for the contribution! Sorry for lack of responses, I think this is great, albeit a bit niche-y in terms of general API's, By that i mean, we are trying to balance the default size of yup with including robust validations and we don't really have a good solution right now for these sort of validations that are useful, but maybe not to everyone. I'm gonna leave this open, I'd really like to find a good option, for including it, maybe like a yup/addons sort of thing but need more time to to think that through

PR #717, opened 2019-12-07
Adds moreThan and lessThan to date validation (#1)

gburnett:
While using this library I have found that it is useful to be able to validate dates with `moreThan` and `lessThan`.

Do you think that this would be a useful addition to the api?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ json separate from validating it, via the `cast` method.
   - [date](#date)
     - [`date.min(limit: Date | string | Ref, message?: string | function): Schema`](#dateminlimit-date--string--ref-message-string--function-schema)
     - [`date.max(limit: Date | string | Ref, message?: string | function): Schema`](#datemaxlimit-date--string--ref-message-string--function-schema)
+    - [`date.lessThan(limit: Date | string | Ref, message?: string | function): Schema`](#datelessthanlimit-date--string--ref-message-string--function-schema)
+    - [`date.moreThan(limit: Date | string | Ref, message?: string | function): Schema`](#datemorethanlimit-date--string--ref-message-string--function-schema)
   - [array](#array)
     - [`array.of(type: Schema): Schema`](#arrayoftype-schema-schema)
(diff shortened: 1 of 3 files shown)


gburnett:
@akmjenkins Thanks, I hadn't noticed `yup.addMethod`.  I think that will probably work for me.  I'll try it and if it works then I'll come back and close this.

I don't think that `min` and `max` will work.  I have a schema with two dates.  The second date has to be the greater than the first.  That is to say that it has to be the following date.  If I use `min`, `max` and pass in refs then it is valid for both dates to be the same date.

the-daniel-rothig:
I agree that yup should be very selective with the validations `yup` ships with by default. **However**, I think `date.lessThan` and `date.moreThan` are sensible additions to the core, simply because `number` has those, and consistency is good.

the-daniel-rothig:
@akmjenkins Yes I agree the argument extends to `array` and `string`. Not sure if that amounts to a slippery slope towards continual additions. 

I imagine that an `inclusive` flag would generate the same amount of code, documentation and test cases as a separate method; but really I am trying to make a point of consistency. I don't know if it is sane to have a `moreThan` and `lessThan` pair in addition to `max` and `min` in the core; but I feel that the decision should be the same across all schema types.

jquense:
Thanks the for the PR, I'd like to have consistent validation options across schema types while also weighing the cost of adding more by default to the core library...I'm still thinking about the best way to manage this long term, and have even considered some fairly drastic api changes like RXJS to enable importing only what you need, e.g.

```js
import { lessThan, required, string } from 'yup'

const schema = string(lessThan(), required())
```

But i've not not reached any conclusions yet. Lets leave this open a bit and see if there is a good way to get it included

Svish:
Wouldn't be much more sensible if what was added to `date` was `after` and `before`, rather than `lessThan` and `moreThan`? That would, in a way, also remove the feeling that they should also be added to other things like `string` and `array`, I think?

PR #727, opened 2019-12-20
Mention the special role of `undefined` in `mixed.oneOf`

mhienle:
Mention that `undefined` does not fail the `oneOf` validator.


jquense:
thanks!

Issue #311, opened 2018-09-14
Publish ESM modules


alexlafroscia:
I'd love to consume this package as an ESM module, so that I can only import the validator functions that I need.

I see that a Rollup configuration is present, but it doesn't seem to actually be used -- the `dist` folder with the CJS and ESM builds is not published. The only thing published is a `lib` directory that the Babel CLI produces.

Is there a reason for this? If not, I would be happy to update the build scripts, etc. to publish the `dist` directory that Rollup produces instead.

jquense:
I just haven't finished the work to do it yet. Happy to take a contribution adding it

alexlafroscia:
Do you have any thoughts on publishing the `esm` build from Rollup vs. just publishing the original source as the ESM build? I'm not sure there's even a benefit having an `esm` consume not import the original modules directly.

stramel:
Related to #221

ReasonX7:
Any updates on it?

jquense:
nope

lundn:
Is this going to happen?

jquense:
maybe!

Issue #100, opened 2017-07-04
Add UMD build


jaredpalmer:
Selfishly want this for forthcoming Formik docs site.

jaredpalmer:
@jquense would you merge a PR that updates tooling with Rollup and adds this?

jquense:
Sure!

ravilution:
I need this badly as well. Please build a UMD that can be accessed via https://unpkg.com/.

macklinu:
Some initial thoughts on resolving this issue:

* Update `rollup.config.js` to output a UMD bundle:

```diff
diff --git a/rollup.config.js b/rollup.config.js
index 442db4e..ae8c725 100644
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -38,6 +38,11 @@ module.exports = [
         file: 'dist/yup.esm.js',
         format: 'es',
       },
+      {
+        file: 'dist/yup.umd.js',
+        format: 'umd',
+        name: 'Yup',
+      },
     ],
     plugins: [...base.plugins, filesize()],
   },

```

* Update `package.json#main` to point to the UMD bundle instead of the CommonJS bundle (possibly?)
* Update the `build` npm script to run Rollup in some way

What else might be involved here? Interested in helping resolve this.

mikabytes:
Why was this issue closed? We're also looking for a UMD build of Yup, and there seem to be people interested in helping out adding support for this?

lasseborly:
This is still very much relevant :+1:

1Map:
5 Years later, no umd

jquense:
there will never be a UMD build at this point it's 2025 and bundlers and native modules have been the standard for nearly a decade. If you want to drop yup into a file in a browser, you can import it lots of ESM registries

PR #738, opened 2020-01-06
Add ESM build

sheepsteak:
Related to #311.

I see that there is a `rollup.config.js` that isn't hooked up yet. I guess that's probably the better way to go long term since people want a UMD build too (#100).

I did experiment and got Babel to output UMD too with another `.babelrc` entry for `"modules": "umd"` but it was multiple files so I'm not sure if that's what's wanted for UMD? I don't use UMD myself so I have no idea.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "release": "rollout",
     "build": "yarn build:commonjs && yarn build:modules && npm run toc",
     "build:commonjs": "babel src --out-dir lib --delete-dir-on-start",
-    "build:modules": "BABEL_ENV=modules babel src --out-dir es --delete-dir-on-start",
+    "build:modules": "babel src --out-dir es --delete-dir-on-start --env-name modules",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,10 +14,13 @@
     "precommit": "lint-staged",
     "toc": "doctoc README.md --github",
     "release": "rollout",
-    "build": "babel src --out-dir lib --delete-dir-on-start && npm run toc",
+    "build": "yarn build:commonjs && yarn build:modules && npm run toc",
+    "build:commonjs": "babel src --out-dir lib --delete-dir-on-start",
+    "build:modules": "BABEL_ENV=modules babel src --out-dir es --delete-dir-on-start",
```
jquense: can we use `--envName esm` instead of the env var, it's a bit more cross platform

sheepsteak: Good catch! Done.

jquense:
thanks!

PR #751, opened 2020-01-17
Use RFC 5322 compliant email regex + unit tests (#507)

triqi:
Original regex was inspired from [here](https://emailregex.com). I've added the length constraints - max 64 for local and max 255 for the domain.


bjarnij:
Both the current and the proposed regular expression do allow unsupported email addresses like:
`áéýúíóþæðö@example.com`
But the following regular expression from https://emailregex.com/ does not:
`(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])`

I briefly tried the email addresses from the unit tests in the PR and there seem to be few email address that give different validation result:
* `" "@example.org`
* `A@b@c@example.com`
* `A@b@c@example.com..au`
* `a"b(c)d,e:f;g<h>i[j\\k]l@example.com`
* `just"not"right@example.com`
* `this is"notallowed@example.com`
* `this still"not\\allowed@example.com`
*`1234567890123456789012345678901234567890123456789012345678901234+x@example.com`

...so neither are perfect

jquense:
All email regex's will be imperfect, it's just name of the game unfortunately. The only truly correct way to validate an email address is to send an email to it and have them confirm it was received. Overall I think that we are going to side on being more broad and less complex on email validation in the future, likely just: "Has an `@` in it and a `.` after it`

paleite:
> All email regex's will be imperfect, it's just name of the game unfortunately. The only truly correct way to validate an email address is to send an email to it and have them confirm it was received. Overall I think that we are going to side on being more broad and less complex on email validation in the future, likely just: "Has an `@` in it and a `.` after it`

Formally speaking, even that is too strict;
user@localhost is a completely valid email address and contains no dot.

PR #756, opened 2020-01-19
fix: use lodash-es instead of regular lodash for ES module builds

astei:
I've just now started using `yup` in combination with `formik`, and so far I like it. However, when I reviewed my bundle size (as I often to do ensure that our application is as speedy as possible) I noticed that `lodash` was getting included twice (once in its ES module form from `formik` and once again from `yup` in CommonJS form). Here's a screenshot of what I mean:

screenshot

After this PR (and switching to the ES module version of the library):

image

In order to achieve this I had to convert the Babel configuration into JavaScript, but this seems to work fine and reduces the bundle size quite nicely.

Verified on browser (Chromium 79.0.3945.130 and Firefox 72.0.1) and Node.js 12.14.0.


babel.config.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+  },
+};
+
+module.exports = function(api) {
+  var configCopy = Object.assign({}, baseConfig);
+  if (api.env('modules')) {
+    configCopy.plugins = [
+      [
+        'transform-rename-import',
```
TrySound: You can add this plugin to env: { modules: { ... }} in base config

jaredpalmer:
Ping @jquense this would be great for formik for minimizing bundle size

jquense:
thanks!

PR #793, opened 2020-03-04
Update url regex

crazorin:
Scenario: use yup to validate URL on a form. Our target audience wouldn't think to add the `http://` in front of `github.com` when entering URL.

Expected behavior: `github.com` is accepted as a valid URL.
Current behavior: URL validation fails for `github.com`.

Proposed solution: in the URL regex make the `//` optional by updating the `rUrl` var in the `string.js` file. 
From: `/^((https?|ftp):)?\/\/(((([a-z]|...` to `/^((https?|ftp):)?(\/\/)?(((([a-z]|`


jquense:
Hey there thanks for the PR, I'm not sure this is a good idea given it's very likely to produce false positives for ppl, effectively passing for any string with a `.` in it. Might make sense to accept an option for missing the protocol, but it'd likely be easier to just use a less specific regex for your usecase via `match` and/or `addMethod`

JaneJeon:
@crazorin one thing I do is to pass URLs through something like https://www.npmjs.com/package/normalize-url through transformation *before* you even begin to validate! That way, you can ensure that the http/https protocol is always present!

PR #805, opened 2020-03-07
Why does InferType() not default to nonRequired()

mauricedb:
Added an explanation as this question about `InferType()` has come up several times and has been reported as an issue with the type definition.


README.md
```diff
@@ -1346,3 +1346,16 @@ const fullPerson: Person = {
...
+### TypeScript setting
+
+For `yup.InferType<T>` to work correctly with required and nullable types you have to set `strict: true` or `strictNullChecks: true` in your tsconfig.json.
+
+### Why does InferType not default to nonRequired()?
+
+This was considered when inplementing `InferType<T>` but decided against.
+
+First of all 
+The semantics of a required property in Yup is not the same as in TypeScript. For example a `Yup.array().of(Yup.string()).required()` will fail validation if you pass in an empty array. A required array should not be empty: [empty arrays are also considered 'missing' values](#arrayrequiredmessage-string--function-schema). The same is true for a `Yup.string().required()` where passing in an empty string `""` is considered invalid while an not empry string `"hello"` is considered valid. With TypeScript both would satisfy required. Another example of that is a `Yup.date()` that will pass validation if you use the string `"2020-02-14T07:52:25.495Z"` if you don't call `.strict()`.
```
jquense, line 1359:
```suggestion
The semantics of a required property in Yup is not the same as in TypeScript. For example a `Yup.array().of(Yup.string()).required()` will fail validation if you pass in an empty array. A required array should not be empty: [empty arrays are also considered 'missing' values](#arrayrequiredmessage-string--function-schema). The same is true for a `Yup.string().required()` where passing in an empty string `""` is considered invalid while the non-empty string: `"hello"` is considered valid. With TypeScript both would satisfy required. Another example of that is a `Yup.date()` that will pass validation if you use the string `"2020-02-14T07:52:25.495Z"` if you don't call `.strict()`.
```

README.md
```diff
@@ -1346,3 +1346,16 @@ const fullPerson: Person = {
...
+
+### TypeScript setting
+
+For `yup.InferType<T>` to work correctly with required and nullable types you have to set `strict: true` or `strictNullChecks: true` in your tsconfig.json.
+
+### Why does InferType not default to nonRequired()?
+
+This was considered when inplementing `InferType<T>` but decided against.
+
+First of all
```
jquense, line 1358:
```suggestion
```

jquense:
awesome thanks. I really appreciate your work @mauricedb

PR #806, opened 2020-03-07
Removed Date.parse shim

RShergold:
The `Date.parse` shim does not support 1 & 2 digit years. 

This proposal simply removes the shim. This will result in dropping support for IE8.


jquense:
Happy to drop code when possible, but the shim here was more about supporting non-conformant ISO strings than it was supporting ie8. I'm a bit concerned that relying on browser implementations is going to give inconsistent results across Browsers and runtimes. That said I'm not an expert on how conformant Date.parse is across runtimes

Issue #224, opened 2018-05-23
Validate localhost url


axedre:
Could you please provide a way to allow localhost as a valid url string? Something like validate.js's `allowLocal` flag: https://validatejs.org/#validators-url:
```
let validationSchema = Yup.object().shape({
    url: Yup.string().url('Please enter a valid URL, including scheme', { allowLocal: true })
});
```
Thanks.

tomalexhughes:
This would be incredibly useful, my application requires setting OAuth redirect URL's within a form that I'm validating. For development these URL's need to be http://localhost:8080 but this does not validate.

Any chance of this @jquense

jquense:
Happy to take a PR for this if someone wants to send it out 👍

edqallen:
I don't have time right this second to make and test the modification and submit a PR, though I might in the next few days.  In case I don't get to it, here's a regex you can use to replace the one `rUrl` one you're already using in `string.js`:

    ^(?:([a-z0-9+.-]+):\/\/)(?:\S+(?::\S*)?@)?(?:(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*\.?)(?::\d{2,5})?(?:[/?#]\S*)?$

I tested it here: https://www.regexpal.com/?fam=104762

It's derived from https://gist.github.com/dperini/729294 with a few modifications:
 - Scheme restriction removed.  Any RFC valid scheme is OK now, not just http(s) and ftp.
 - Private IP restriction removed. (i.e. 192.168.1.2... is OK now)
 - TLD requirement removed. ('localhost' and other single-word hostnames OK now)

dperini's regex is pretty widely regarded as the best one for matching URLs.  The alternative is to use a real URL parsing lib instead of a regex, which is probably a good idea, but will increase the size of this lib.

diogommartins:
> I don't have time right this second to make and test the modification and submit a PR, though I might in the next few days. In case I don't get to it, here's a regex you can use to replace the one `rUrl` one you're already using in `string.js`:
> 
> ```
[...]

PR #844, opened 2020-04-17
Allow localhost in string.url()

mlogan:
I've seen a few requests for this, and just ran into this issue myself. Thanks for taking a look.

(Its hard to tell from the diffs, but all I did was add a `localhost|` alternation in the regex.)


maxifom:
why not follow this regexp of RFC3986? Or current regexp does support more cases?
https://gist.github.com/curtisz/11139b2cfcaef4a261e0

micha149:
Why not using the regex @edqallen posted in #224? For example the restriction on http(s) makes no sense for me.

dbouclier:
it's look lile it's solve in https://github.com/jquense/yup/pull/1859
@jquense any update on this ?

PR #850, opened 2020-04-21
Allow passing of function to .matches() options/message param

### LINKED ISSUE #849: Error message provided via function not being returned in validation errors when using .matches()
**Describe the bug**
As of version 0.28.4 the default error message is returned when a function is provided for the message param in string.matches() rather than the given message within the function

**To Reproduce**
[https://codesandbox.io/s/eager-leftpad-sbpxk?file=/src/index.test.js](https://codesandbox.io/s/eager-leftpad-sbpxk?file=/src/index.test.js)

**Expected behavior**
As per prior versions, expect to pass a function that returns a string and be provided with the given message in the event of an error

 - Browser Chrome
 - Version 80

**Additional context**
Aim to provide a PR to fix this

KeelanM90:
Came across issue #849 this morning. Hopefully this closes #849

### DIFF
--- a/src/string.js
+++ b/src/string.js
@@ -77,6 +77,7 @@ inherits(StringSchema, MixedSchema, {
 
     if (options) {
       if (typeof options === 'string') message = options;
+      if (typeof options === 'function') message = options;
       if (typeof options === 'object') {
         ({ excludeEmptyString, message, name } = options);
       }
--- a/src/string.js
+++ b/src/string.js
@@ -76,10 +76,10 @@ inherits(StringSchema, MixedSchema, {
     let name;
 
     if (options) {
-      if (typeof options === 'string') message = options;
-      if (typeof options === 'function') message = options;
       if (typeof options === 'object') {
         ({ excludeEmptyString, message, name } = options);
+      } else {
+        message = options;
       }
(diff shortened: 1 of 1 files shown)


KeelanM90:
Please let me know if any updates are needed

src/string.js
```diff
@@ -77,6 +77,7 @@ inherits(StringSchema, MixedSchema, {
 
     if (options) {
       if (typeof options === 'string') message = options;
+      if (typeof options === 'function') message = options;
```
jquense, line 80: I think this whole inner branch can be rewritten as 
```js
if (typeof options === 'object') {
  ({ excludeEmptyString, message, name } = options);
} else {
  message = options
}
```

KeelanM90: Thanks for the review, I had defaulted to checking it was a function as there was type checking for strings. The docs do show the supported types so it's really down to user error if they pass an incompatible type

jquense: in practice, messages can be any type! The string check seems like a bug

KeelanM90: Perfect, have pushed the update, thanks again!

jquense:
thanks!

PR #871, opened 2020-05-02
Bug: fix object bug when nested object has a property with strict

### LINKED ISSUE #857: Can't use strict in nested objects.
**Describe the bug**
When I try to use strict in a nested object nothing happens, e.g

```js
const schema = yup.object({
    obj: yup.object({
      name: yup.string().strict()
    })
  });
```

`name: 131231` and `name: {}` will return as a valid schema.

**To Reproduce**

https://codesandbox.io/s/wonderful-night-ci7cu?file=/src/index.test.js

The first test is passing but should fail.

**Expected behavior**
Should return false

 - Version 0.28.4

netochaves:
Closes #857 

This PR fix a bug when `strict` is used on nested objects and the cast still happens

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -88,7 +88,7 @@ inherits(ObjectSchema, MixedSchema, {
     let innerOptions = {
       ...options,
       parent: intermediateValue,
-      __validating: false,
+      __validating: options.__validating || false,
     };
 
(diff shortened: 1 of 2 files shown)


jquense:
thanks!

PR #885, opened 2020-05-10
Expose blacklist and whitelist values on description

### LINKED ISSUE #883: scheme.describe omits whitelist and blacklist
**Describe the bug**
[schema.describe](https://github.com/jquense/yup/blob/master/src/mixed.js#L532-L542) does not expose the whitelist or blacklists maintained by [oneOf](https://github.com/jquense/yup/blob/master/src/mixed.js#L470-L471)

**To Reproduce**
https://codesandbox.io/s/nice-wilson-egh66?file=/src/index.test.js
```js
const yup = require("yup");
const assert = require("assert");

it("should pass", async () => {
  const schema = yup.object({
    name: yup.string().oneOf(["a", "b", "c"])
  });

  const description = schema.describe();
  assert.deepEqual(description.whitelist, ["a", "b", "c"]);
});
```

**Expected behavior**
The whitelist and blacklists to be accessible (as either sets or arrays) through schema.describe();

**Additional context**
I would like to use this to generate test data for [yup-by-example](https://github.com/cressie176/yup-by-example). I would be happy to submit a PR.

dreamsinspace:
Closes #883

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -550,8 +550,8 @@ const proto = (SchemaType.prototype = {
         ),
     }
 
-    if (next._whitelist.size) description.whitelist = next._whitelist.describe();
-    if (next._blacklist.size) description.blacklist = next._blacklist.describe();
+    if (next._whitelist.size) description.oneOf = next._whitelist.describe();
+    if (next._blacklist.size) description.notOneOf = next._blacklist.describe();
     
     return description;
   },
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -17,6 +17,17 @@ class RefSet {
     this.list = new Set();
     this.refs = new Map();
   }
+  get size() {
+    return this.list.size + this.refs.size;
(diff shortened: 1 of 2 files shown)


src/mixed.js
```diff
@@ -538,7 +548,12 @@ const proto = (SchemaType.prototype = {
         .filter(
           (n, idx, list) => list.findIndex(c => c.name === n.name) === idx,
         ),
-    };
+    }
+
+    if (next._whitelist.size) description.whitelist = next._whitelist.describe();
```
jquense, line 553: lets call these `oneOf` and `notOneOf` to match the method name, it'll involve a little less mental overhead mapping schema to descriptions etc

dreamsinspace: Sure thing @jquense

dreamsinspace:
Np, glad it gets merged.

Issue #904, opened 2020-05-21
Provide a first-class support for checking truthiness of booleans


tomdohnal:
Hi :)

I've been enjoying using yup to validate both forms and API requests, so I'd like to say a huge thank you in the first place! :) 

Anyways, what I'm missing is a more canonical way of checking if a boolean is `true` that doing something like: 
```javascript
yup.boolean().oneOf([true]),
```

Although the above-mentioned snippet does the job perfectly, it *feels* to me like a workaround for missing functionality. And I *believe* it's quite common to require a boolean to be `true` (checking checkboxes with agreements, ...).

I know that changing how `.required()` works for booleans (to only accept `true`) values is not possible (backwards compatibility, ...). 

What I'd suggest is to either create a general `truthy()` checker on the `mixed()` type which would check if the value passed is truthy as JavaScript evaluates it (e. g. everything except for `null`, `undefined`, `0`, `false` and some other). 

Or create an explicit `isTrue()` checked on the `boolean()` type which would simply check if the value is `true`. 

If you'd be interested in adding this functionality to the API, I'd be happy to send a PR with whatever solution you'd prefer :)

jquense:
I'd be ok with a isTrue and isFalse methods on boolean

tomdohnal:
OK, thank you. I'll do my best to send a PR within a week :)

PR #910, opened 2020-05-25
add `isTrue` and `isFalse` checks on boolean

tomdohnal:
This resolves https://github.com/jquense/yup/issues/904 :)

### DIFF
--- a/src/boolean.js
+++ b/src/boolean.js
@@ -1,5 +1,7 @@
 import inherits from './util/inherits';
 import MixedSchema from './mixed';
+import { boolean as locale } from './locale';
+import isAbsent from './util/isAbsent';
 
 export default BooleanSchema;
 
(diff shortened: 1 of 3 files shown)


src/boolean.js
```diff
@@ -25,4 +27,26 @@ inherits(BooleanSchema, MixedSchema, {
...
+      test(value) {
+        return isAbsent(value) || value === true;
+      },
+    });
+  },
+
+  isFalse(message = locale.isFalse) {
+    return this.test({
+      message,
+      name: 'isFalse',
```
jquense, line 45: the `exclusive` flag doesn't do anything unless there is another test with the same name. So you want to name both of this something like `is-value`, then `bool().isFalse().isTrue()` will only use the last isTrue, instead of creating an unmatchable schema

src/locale.js
```diff
@@ -51,7 +51,10 @@ export let date = {
   max: '${path} field must be at earlier than ${max}',
 };
 
-export let boolean = {};
+export let boolean = {
+  isTrue: '${path} field must be true',
+  isFalse: '${path} field must be false',
```
jquense, line 56: you can get away with one message here: `isValue: '${path} field must be ${expected}', by adding the expected value to the test `params`

src/boolean.js
tomdohnal: Oh, I see. I updated it according to your suggestion

src/locale.js
tomdohnal: Okay, done : )

lucasff:
I think this PR is good to go. Let's merge it?

jquense (approved):

PR #936, opened 2020-06-11
feat: export locale

fjc0k:
This is useful for custom method. eg:

```js
import * as yup from 'yup'
import { isChineseMobilePhoneNumber } from '../utils'

yup.setLocale({
  string: {
    chineseMobilePhoneNumber: '${path} must be a Chinese mobile phone number'
  }
})

yup.addMethod(
  yup.string,
  'chineseMobilePhoneNumber',
  function (message = yup.locale.string.chineseMobilePhoneNumber) {
    return this.test(
      'chineseMobilePhoneNumber',
      message,
      isChineseMobilePhoneNumber
    )
  }
)
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,6 +11,7 @@ import ValidationError from './ValidationError';
 import reach from './util/reach';
 import isSchema from './util/isSchema';
 import setLocale from './setLocale';
+import locale from './locale';
 
 let boolean = bool;
 let ref = (key, options) => new Ref(key, options);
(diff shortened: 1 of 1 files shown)


tpict:
You can import this from `yup/lib/locale`, but that causes issues for TypeScript users because the `@types/yup` declaration file won't be applied to that module. This PR would fix that issue!

knicola:
we can now use `defaultLocale` ie. `yup.defaultLocale.string` though the internal locale types, such as `StringLocale`, are not exported so declaration merging is not working.

Issue #676, opened 2019-10-26
TypeError: Cannot read property 'fulfilled' of undefined


thoughtassassin:
**Describe the bug**
I have a Codesandbox running a library called React Hook Form. If you add two friends in the code pen and then remove the first friend and then submit the form you see the error. The stack trace shows the error at a line in react-hook-form.js. BUT if you dig deeper the error is actually coming from line 84 of runValidations.js where the check for !r.fulfilled throws an error saying:

`"TypeError: Cannot read property 'fulfilled' of undefined
    at eval (eval at <anonymous> (https://yycrz.csb.app/node_modules/yup/lib/util/runValidations.js), <anonymous>:1:3)
    at eval (https://yycrz.csb.app/node_modules/yup/lib/util/runValidations.js:84:17)
    at Array.filter (<anonymous>)
    at eval (https://yycrz.csb.app/node_modules/yup/lib/util/runValidations.js:83:32)
    at async Promise.all (index 0)
    at async validateWithSchema (https://yycrz.csb.app/node_modules/react-hook-form/dist/react-hook-form.js:383:21)
    at async eval (https://yycrz.csb.app/node_modules/react-hook-form/dist/react-hook-form.js:886:28)"`

**To Reproduce**
https://codesandbox.io/s/react-hook-form-array-fields-validation-using-refs-yycrz?from-embed

Add two friends and then delete one friend. When you submit the form, you can see the error Chrome dev tools.

> NOTE: if you do not provide a runnable reproduction the chances of getting feedback are significantly lower

**Expected behavior**
Handle the scenario where r is undefined to fail gracefully so it does not break the running script.

**Platform (please complete the following information):**

**Additional context**
Also, one of the folks at React Hook Form asked me to share this screenshot:
<img width="730" alt="1b0f20e8-8e5b-4b01-ac56-5ab844645213-Screen%20Shot%202019-10-26%20at%208 52 04%20am (1)" src="https://user-images.githubusercontent.com/1539057/67612988-b879d980-f76d-11e9-9a2d-39463cda6200.png">

sijad:
I faced same issue:

```
Unhandled Rejection (TypeError): r is undefined
```

which refer to this line: https://github.com/jquense/yup/blob/0c27f076668632912311226259745d19129bf9bb/src/util/runValidations.js#L62

VanTanev:
[...]

PR #949, opened 2020-06-24
Handle sparse array positions as undefined

VanTanev:
This fixes https://github.com/jquense/yup/issues/676

---

The is the source of the bug is here, but rather than adding a local fix, a better fix is proposed in this PR:

https://github.com/jquense/yup/blob/27b287ba0dd8981ddb330c28123b41d85c4c8853/src/util/runValidations.js#L38-L49

Here, `promises.map` is supposed to map fulfilled and rejected promises to the appropriate value object. However, `Array.map` does not iterate over the empty spots in sparse arrays:

image

The bug originates here:

https://github.com/jquense/yup/blob/27b287ba0dd8981ddb330c28123b41d85c4c8853/src/array.js#L87

Here, no validation will be generated for empty Array positions.

---

It seems reasonable for yup to transform empty array positions to undefined in its output. This fixes the bug, and is arguably what people expect when a sparse array is passed into yup.

### DIFF
--- a/src/array.js
+++ b/src/array.js
@@ -84,6 +84,12 @@ inherits(ArraySchema, MixedSchema, {
 
         originalValue = originalValue || value;
 
+        // Populate sparse arrays positions with undefined
+        value = value.slice()
+        let len = value.length;
+        let idx = -1;
+        while (++idx < len) if (!(idx in value)) value[idx] = undefined;
+
         let validations = value.map((item, idx) => {
           var path = makePath`${options.path}[${idx}]`;
 
--- a/test/array.js
+++ b/test/array.js
@@ -202,4 +202,14 @@ describe('Array types', () => {
       .of(itemSchema)
       .validate(value);
   });
+
+  it('should treat sparse arrays empty spots as undefined', async () => {
+    let sparseArray = new Array(2);
+    sparseArray[1] = 1;
+    let value = await array().of(number()).validate(sparseArray);
+    // assert the sparse array position has been turned into a real position
+    expect(0 in sparseArray).to.be.false()
+    expect(0 in value).to.be.true()
+    value.should.eql([undefined, 1]);
+  });
 });


src/array.js
```diff
@@ -84,6 +84,12 @@ inherits(ArraySchema, MixedSchema, {
 
         originalValue = originalValue || value;
 
+        // Populate sparse arrays positions with undefined
+        value = value.slice()
+        let len = value.length;
+        let idx = -1;
+        while (++idx < len) if (!(idx in value)) value[idx] = undefined;
```
jquense, line 91: Thanks for working on this! This does solve the problem, but in doing does a lot more work in a relatively hot path. Before there was a single array iteration but now the array is being cloned and looped over another time. 

Supporting sparse arrays is fine insofar as it doesn't cost much since they are not heavily used and often represent bugs more than a desire to use a real sparse array. In any case a more surgical fix here would be to change `runValidations` allSettled function to check that the result is defined before checking if it was fulfilled.

VanTanev: Making the change there does not output `undefined` in the empty positions of `schema.validate(sparseArray)`, but keeps them empty. If we're OK with that, there is a simpler change to be made here, it looks like this:

https://github.com/jquense/yup/compare/master...VanTanev:handle-sparse-arrays-no-fill-undefined

VanTanev: Note that if we had `y.array().of(y.number().required()).validate(sparseArray)`, Adding the fix in runValidations would errorneously miss the sparse position being required

jquense: I am ok with sparse entries being skipped, since that's basic the native behavior. It's weird, but i don't think it's yup's job to make choices for how to "fix" sparse arrays, in practice the developer should not use them if they are not comfortable with how they work.

Quick note tho, i don't your linked commit is as simple as i'm proposing, `array.map` _already_ keeps the empty slots empty, the fix would be to address the actual site of the error here which is `allSettled` assuming the array is not sparse when it filters

VanTanev: Sorry, I don't think I've communicated clearly enough, let me try again:

1) `Array.map()` of a sparse array produces a new sparse array
2) The validations for each item of an array are generated by https://github.com/jquense/yup/blob/27b287ba0dd8981ddb330c28123b41d85c4c8853/src/array.js#L87

Because of that, `validations` itself becomes a sparse array, and that's an issue when the schema defines array members to be required. With a fix in `runValidations.js` instead of `array.js`, you will get the following bug:

```javascript
let arr = Array(2)
arr[1] = 1
let schema = y.array().of(y.number().required())

schema.isValid(arr)
// => true, but it should be false
```

jquense: I think the disconnect here is I don't think `array(number().required()).isValid(new Array(2)) -> true` is necessarily a bug. the empty slots in sparse arrays are holes in the Array, not an element of it. E.g. they aren't undefined values. You can definitely make the argument that coercing holes to `undefined` is a good idea, but i don't it's worth it for a few reasons. 

- Sparse arrays are an edge case and shouldn't be used for these sort of cases anyway. In the same way yup doesn't deal explicitly with Functions, Regexps, or Boxed primitives directly, schema are for validating more primitive, generally serializable data.
- If someone is _intentionally_ using sparse arrays, it's likely they are doing so because of the special properties empty slots so it'd be preferable to maintain it.

VanTanev: I have to disagree.

The most common way to get a sparse array is probably manually assigning to a index pass the current array length.

image

Indeed, this is the source of the bugs with `Formik` and `react-hook-form` in #676. 

More generally, yup's goal is validation - if I say "I want an array of numbers", and yup treats an array full of holes as an array of numbers, this is highly unexpected, and highly undesirable, in my eyes.

VanTanev: btw, do note that the proposed fix keeps sparse arrays sparse, if that's allowed by the validation: https://github.com/jquense/yup/compare/master...VanTanev:handle-sparse-arrays-no-fill-undefined#diff-bba3cb87d6077518c25974be601ab17bR210-R211

VanTanev: Closing in favor of #950

VanTanev:
Closing in favor of #950

src/array.js
jquense: This is definitely a bug that arises, but I think it proves my point about SA's being weird language edge cases and their existence is accidental or actively a bug. e.g. formik and RHF should not be creating sparse arrays that is a bug on their end if they are. 

The approach you are advocating is totally valid, as I said, I don't think it's the one we should take for yup generally. Sparse arrays are actively different from arrays filled with a mix of values and `undefined`, which is one reason why the console representation of them isn't `[undefined, 1]`. Changing the value to `undefined` is an act of coercion on the input. It's a reasonable one but it's not appropriate to do in `validate` given how `yup` has delegated the duties of transformation and validation. `validate` should/must validate value _as is_. Coercion needs to be done in `cast` and apart from the initial basic type coercion should be opt-in. As you say if you want to handle SA's then `compact` already exists for that purpose

VanTanev: I have elaborated on the behavior in #950

In my PR, the output of sparse array validation is still a sparse array, so no yup doesn't do coercion.

It just applies validation to empty slots as well, instead of ignoring them. Ignoring empty slots and not applying validation to them is, I think, highly unexpected.

There are valid reasons why `Formik` and `react-hook-form` create sparse arrays - if you have a list of 5 fields that map to a tuple, represented as an array, then it very much makes sense to have that array be sparse if a user starts entering in the third field.

More generally, the reason this happens is because both those libs (and other similar ones) set value on an object using a path function, eg `user.friends[3].name`. Such setter functions always have the potential to produce sparse arrays. 

Note that if we're dealing with a tuple, compact() is completely invalid way to handle things. It will move the error message to the wrong index:
```javascript
let data = []
let data[1] = ['short']

let schema = array(string().min(10))
schema.compact().validate(data) // error is on index 0, but it should be on index 1. 
```

VanTanev: Hey, I know I'm getting a bit too insistent, so this will be my final argument. The ultimate decision will be up to you.

I just wanted to re-iterate, the change proposed in #950 does not cast the output of sparse arrays - they're preserved in the output of `schema.validate()`. If a sparse array passes validation, you get a sparse array as output.

What the change does is, it applies the specified inner array validation even to the empty slots in sparse arrays.

To me, this seems like the expected behavior - inner validation specification should be applied to all member of the array, even empty ones. 

There is inconsistency in how similar things are handled in the language itself. While `Array.prototype.map()` and `Array.prototype.forEach()` do not iterate over empty positions,  `for..of` / `arr[Symbol.iterator]()` do iterate over them, and `(new Array(2)).includes(undefined)` returns true for sparse arrays.

It's also worth noting that validating empty positions is not a breaking change in Yup. Previous behavior was the bug in #676 - ie, throw an exception.

Overall, it is my belief that applying the given schema validation to empty positions makes the most sense. I would expect a schema like `array(string().defined())` to make sure I have no empty positions in the array under validation, while `array(string())` to allow and pass them through.

Thank you for your patience, and I apologize if I have been overly insistent in my communication.

jquense: No worries, and thanks for checking your tone (I also apologize if i come off as dismissive). This isn't my job and can get tiring to debate folks about this stuff over and over.

**tl;dr;** from me is that i still think this is unconvincing and bad reason to use sparse arrays, however, you've convinced me that folks probably don't realize they are creating them and likely intend the empty slots to be validated.

Slightly longer version is that, i've written a lot of form libraries (and a validation library for them ;) ) with complex tuple-like fields and haven't run into this issue. I think it's a bad idea to build ui not backed by an actual filled value for each item in the array, and that relying on `_.set` functions to build the array is also a bad idea. That said, it seems like if someone is constructing Fields from a `while 0 to 5` loop (why tho??) they won't necessarily realize that it constructs sparse arrays or why those arrays are weird. I still think it will more often than not bite ppl but they probably do mean to count the empty slots as actual items, e.g. an array of 5 undefined values vs an array with one value and a length of 5.

VanTanev: Thank you for reconsidering!

Rest of this comment describes technical details of how I hit #676, feel free to ignore if you're not interested :)

I had a situation like:
```tsx
// API object
type Relationship = {
  characters: [Character, Character] // tuple of 2 characters
}
type Character = {
  id: string
  name: string
}

const RelationshipSchema = y.object({
  characters: y.array()
    .of(y.object({ id: y.string().required(), name: y.string().required(), }).required())
    .required()
    .min(2)
    .max(2)
})

// in actual code this was type RelationshipFormValue = y.InferType<typeof RelationshipSchema>
type RelationshipFormValue = {
  characters: Array<Character> // here it's inferred as an array and not a tuple
}

// in the form, something like:
const initialFormValue: RelationshipFormValue = {
  characters: []
}

// and then 2 selects for character 1 and character 2
<SelectCharacter name="characters[0]">
<SelectCharacter name="characters[1]">
```

So, the problem is, I have set `characters` prop to not allow `undefined` as a value. Because of that, my initial array is an empty array, instead of something like `characters: [undefined, undefined]`.

There is the implication that the form value before validation can be kind of anything. Strictly speaking, the actual type of the form value before validation is:
```
type RelationshipFormValue = {
  characters: Array<Character | undefined>
}
```

But converting back and forth between these types is somehow meaningless if the validation library guarantees that we get the proper data out the end.

In any case, the user could fill in the second character dropdown first, which would create a sparse array. Ultimately, at that point, I would like to show an error under the first character dropdown "Character 0 is required." However, because of #676, I ended up adding `.compact()` to the array validation, and showing a different error message: "You must select 2 characters".

Overall, there were ways of working around the issue - initializing with `characters: [undefined, undefined]` and updating the types would have been probably the most appropriate.

PR #961, opened 2020-07-03
README: document interpolation with default values

jgehrcke:
I am new to yup and carefully read the documentation to understand what I need to do to i) validate a user-given object against a schema and also to ii) fill the user-given object with defaults set in the schema.

Based on what's in the README I couldn't explicitly find an answer for how to do ii). I then tried the obvious thing; inspecting the value returned by validation. Yes, that seems to be the way to go.

I think it's worth explicitly discussing this use case in the README. Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,14 @@ schema.cast({
   createdOn: '2014-09-23T19:25:25Z',
 });
 // => { name: 'jimmy', age: 24, createdOn: Date }
+
+// interpolate with default values
+let schema = yup.object().shape({
+  count: yup.number().required().positive().integer(),
+  version: yup.string().default("v1")
+});
+schema.validateSync({ count: 1 });
+// => { version: 'v1', count: 1 }
 ```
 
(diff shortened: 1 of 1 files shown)


jgehrcke:
Thanks for engaging! :)

> This is documented at 

Mh! Before I opened this ticket I did read this section more than once because, of course, _this_ section should answer my question, right? But yeah, I did not find the answer to my question in there. Not in a way that I could understand it. I hope that kind of feedback is valuable.

The question users come in with is, in my opinion *"How can I fill the user-given object with defaults set in the schema?"*

As answers to this question I do understand 

> the return value of `validate()` contains the user-given object filled with defaults set in the schema

 or, in other words, inspired by the working you just used:

>  object fields will have their defaults applied automatically when calling `validate()` (reflected in the return value)

The key information really is the the _return value of the `validate()` class of functions_ is what we need when we're interested in the original object interpolated with defaults.

Issue #797, opened 2020-03-05
typeError() with no param on number field doesn't show any default msg


rretter:
Using the following Yup defn for the checking of a field:

`fieldname: Yup.number().typeError()`

results in no error msg appearing when the field has a non-number value such as 'fargle'.

(As a note: although it's being checked for number-ness, the field has 'type' property = 'text'.)

Now the following:

`fieldname: Yup.number().typeError('this actually appears')`

_does_ show 'this actually appears' when the field has a non-number value.

So it seems necessary to provide a msg explicitly to typeError() but _not_ to other methods, such as positive() or min().  Those default the msg text; for instance, positive() causes "${path} must be a positive number" to appear when the field has value "-3".

Why doesn't typeError() show a default message (or why isn't it appearing)?

illuminist:
https://github.com/jquense/yup/blob/master/src/mixed.js#L447

The function doesn't have default message parameter. Locale file also doesn't have default message for typeError, opposed to  oneOf and notOneOf validator.

jquense:
if you don't include a message, it makes the message undefined, which is likely why you aren't seeing any message. typeError requires a message paramater

rretter:
Thank you for your description of the problem.

But actually, a _solution_ to the problem would be better.

jquense:
the solution is to not call typeError() without a message parameter.

PR #976, opened 2020-07-14
feat: add default message to typeError from locale

brunohkbx:
## Motivation
I couldn't find a way to override `typeError` message using a custom locale dictionary. It only works passing a default message to `typeError` as described here: #797

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -460,7 +460,7 @@ const proto = (SchemaType.prototype = {
     return next;
   },
 
-  typeError(message) {
+  typeError(message = locale.notType) {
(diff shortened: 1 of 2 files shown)


jquense:
hmm the error _is_ already set from the locale: https://github.com/jquense/yup/blob/master/src/mixed.js#L81

I wouldn't have expected this change to affect behavior.

brunohkbx:
@jquense the test that I wrote will break if I remove the default value `typeError(message = locale.notType)`

Maybe this `this.typeError(locale.notType);` isn't working as expected? :thinking:

PR #978, opened 2020-07-15
Add engines in package.json

MichaelDeBoey:


jquense:
yup doesn't support node 8 its EOL, i'm not sure i'd even want to claim it supports node 10.

PR #991, opened 2020-07-29
fix!: strings fail when trying to coerce Arrays or plain objects

jquense:
BREAKING CHANGE: array or object inputs to string schema will fail now, for old behavior add:

```js
string().transform(function (current, original) {
  return this.isType(current) ? current : String(original)
})
```


jquense:
cc @sloria I am not sure the array one is a good idea...arrays do have a reasonable toString (unlike plain objects) so more likely the old behavior was desirable for folks

sloria:
Not sure either. Seems a bit odd to coerce non-scalar values to strings, but I don't feel strongly about it.

sloria:
That said, the current fix I'm working on wouldn't work without the Array handling you have in this PR

src/string.js
```diff
@@ -8,9 +8,10 @@ let rEmail = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\u
...
 // eslint-disable-next-line
-let rUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+let rUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
 let isTrimmed = (value) => isAbsent(value) || value === value.trim();
+let objTag = {}.toString();
```
taion, line 14:
```suggestion
const objTag = {}.toString();
```

src/string.js
```diff
@@ -8,9 +8,10 @@ let rEmail = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\u
...
 // eslint-disable-next-line
-let rUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+let rUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```
taion, line 11:
```suggestion
const rUUID = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
```

src/string.js
```diff
@@ -8,9 +8,10 @@ let rEmail = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\u
...
 
 let isTrimmed = (value) => isAbsent(value) || value === value.trim();
```
taion, line 13:
```suggestion
const isTrimmed = (value) => isAbsent(value) || value === value.trim();
```

src/string.js
```diff
@@ -20,7 +21,12 @@ export default function StringSchema() {
   this.withMutation(() => {
     this.transform(function (value) {
       if (this.isType(value)) return value;
-      return value != null && value.toString ? value.toString() : value;
+      if (Array.isArray(value)) return value;
+
+      const strValue =
+        value != null && value.toString ? value.toString() : value;
+      if (strValue === objTag) return value;
```
taion, line 28: is `typeof` no good here? i guess you'd have to do `instanceof String` as well in case there's a `new String('foo')`?

PR #995, opened 2020-08-01
Add parent access to array schema

pipopotamasu:


jquense:
not sure yet if the current behavior is intentional, need a free moment to think it through. thanks for the PR!

Issue #1007, opened 2020-08-11
Feature Request: originalValue in test


shellscape:
🍻  I'd like to propose making the `test` function a bit more useful for complex tests:

## Feature Use Case

We have the need to assert that a string is a valid date, that it isn't before a particular date, _and_ that it matches a certain format. As such, we've chosen to use `test` for that, and this is what we've come up with:

```
const schema = yup.object().shape({
  untilDate: yup
    .date()
    .min('01-01-2020')
    .test('format', '${path} must match the mm-dd-yyyy format', function () {
      return reDate.test((this.options as any).originalValue);
    }),
});
```
If we use `value` there, it contains a full ISO string value. e.g. `'2020-08-11T23:58:15.712Z'`, and we want to validate what was passed, not what the transformed/translated value is.

The `(this.options as any)` bit there is TypeScript, but it signifies that `originalValue` is not documented on `this.options` within the `test` function. It does exist, but it's not been officially documented. And as #207 suggests, it may not always be present. That makes using it in our scenario a little dicey.

## Feature Proposal

I'd like to propose adding `original` value to the `test` function handler parameters. Nearly identical to the signature for [`transform`](https://github.com/jquense/yup#mixedtransformcurrentvalue-any-originalvalue-any--any-schema), so that the [`test`](https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema) function would resemble:

```
const schema = yup.object().shape({
  untilDate: yup
    .date()
    .min('01-01-2020')
    .test('format', '${path} must match the mm-dd-yyyy format', function (value, originalValue) {
      return reDate.test(originalValue);
    }),
});
```

jquense:
Happy to add the original value somewhere, tho maybe `this` is more appropriate given the current API. I'm reticent to adding new functions args and have generally regretted placing `originalValue` as a second arg in `transform()`. I recognize that `this` is also annoying since it precludes arrow functions, perhaps ideally we also pass `this` as the second argument like we do for `when` conditions

jquense:
Also it's a little weird in cases where `originalValue` is just `value` e.g. casts haven't run, which may be why it's missing sometimes

shellscape:
All of those ideas sound completely reasonable to me!

jquense:
open to a PR if you want to get it done fast, i'm a bit short on time atm

shellscape:
Understood. I'll see about fitting this into our next sprint.

abnersajr:
[...]

Issue #207, opened 2018-04-26
`mixed.test` function, `this.options.originalValue`, should i use it?


gabor:
i'm trying to validate a number-field, and to make sure it only contains a number (the default validator accepts things like "42haha". i did something like this:

```
const schema = yup.object().shape({
    num: yum.number().integer.required().test('foo', 'must-be-num', function(value) {
                    const orig = this.options.originalValue;
                    return value.toString() === orig.toString();
              }),
});
```

the question is, is the `this.options.originalValue` guaranteed to be there? i could not find anything about it in the documentation.

the `this.options` field only says:
```the options object that validate() or isValid() was called with```

`validate`  says:
```
Options = {
  strict: boolean = false;
  abortEarly: boolean = true;
  stripUnknown: boolean = false;
  recursive: boolean = true;
  context: ?object;
}
```

so no info about the `originalValue` thing.

i'd just like to make sure i am using something i am supposed to use, and not some internal detail that can change with the next version.
alternatively, if there is a simpler way to achieve stricter number-validation, or a simpler way to get the `originalValue` in the `test`, please tell me.

shellscape:
@jquense why was this closed without a response? Ran into the same thing today and was searching the issues for an answer.

PR #1021, opened 2020-08-27
Feat: exposes context on mixed.test function and add originalValue to context

abnersajr:
This PR fixes two issues:  #207 and #1007.
The discussions suggest having `originalValue` exposed in some way. Exposing as the second argument was one of the suggestions.
Following the API and suggestion from @jquense, would be better to have this `originalValue` into the  `this` context.

In previous versions, you can't use arrow functions too, because it's being used `.call(ctx...)`.
With these changes, you can use both ways. The `this` context is now exposed too as a second argument.
Eg: `mixed.test('name', 'test name', (value, ctx) => ctx.originalValue)`

Also edited Type Definitios: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47069


README.md
```diff
@@ -706,26 +706,27 @@ use the alternate signature to provide more options (see below):
 let jimmySchema = string().test(
   'is-jimmy',
   '${path} is not Jimmy',
-  value => value === 'jimmy',
+  value, context => value === 'jimmy',
```
jquense, line 709:
```suggestion
  (value, context) => value === 'jimmy',
```

jquense:
Hi @abnersajr thanks for the PR. It looks like this PR updates the docs but doesn't implement the feature. Is that intentional or did a commit go missing somewhere?

abnersajr:
Sorry @jquense. Definitely something was misssed in my commits. I'm gonna check it ASAP. and thanks for the quick review.

alexandernst:
What happened to this?

alexandernst:
@jquense Are you're still willing to merge this? If there is something missing, I can fork this branch and add it.

alexandernst:
@jquense ping

alexandernst:
@jquense ping

alexandernst:
@jquense ping

jquense:
can you resolve the conflict @abnersajr ?

abnersajr:
@jquense sure. I'm gonna do this.

abnersajr:
@jquense merge is done. Solved the conflicts adding to the sync and async version.

jquense:
thanks!

alexandernst:
\o/

can we get a new minor release with this, please? 😄

alexandernst:
@jquense Actually, I think this might require to bump the version up to `0.30.0` instead of `0.29.4`.

Issue #1017, opened 2020-08-26
Performance issues for backend use


JaneJeon:
Hi.

I love the API of yup, and I love it SO much that I'm trying to build an isomorphic application that uses the same yup schema validation in both the frontend and the backend 🚀.

However, at the moment it's not really viable simply due to the fact that it's about one order of magnitude slower than it *should* be, considering what is actually running in the source code ("joi-lite"). I really don't expect `ajv` levels of "extreme" performance since `yup` is not compiling schemas, but I'd expect to at least be somewhere around `joi`'s level.

First, the setup: https://github.com/JaneJeon/validator-benchmark (I forked https://github.com/icebob/validator-benchmark, updated the dependencies, added strict mode to the yup schema, and properly ran the async validator - I tested synchronous version, and it was literally 2x slower xP)

And, the results (run `node .`):
https://i.imgur.com/AuQLnTD.png

At first, I thought it was because the email/url regex yup was using was ungodly (no offense), so I thought maybe using simplified regex (like here: https://github.com/icebob/fastest-validator/blob/master/lib/rules/email.js) might help.

It didn't.

I changed the regex in both `lib/string.js`, `es/string.js` in `node_modules/` and it didn't affect the performance in any noticeable way.

I messed around with sync vs. async, different ways of running the validation, etc.

Then I realized there's babel's crufty code all over the place, in both `es/` and `lib/`. I dug around and found that babel-compiled code often has performance issues, especially regarding ES6 syntaxes.

I suspect that the default babel compilation is what's slowing down your codebase significantly since, as far as I can tell, you're only using babel for ESM syntax.

Is there any way you could directly expose the ESM code or at least use a much more efficient compiler/babel setting? People running this in node/backend just need to convert ESM -> CJS. I feel like babel config is key here.

I'd really love to use this in my backend and bring it to production! This is the last thing that's preventing me.

Thanks

JaneJeon:
Update: I hacked both `yup` (which was a pain since ESM doesn't support `import foo from './blah'` and `import { getter } from 'property-expr'` syntaxes) and `validator-benchmark` (and its dependent library, `benchmarkify`) to use strictly ESM. As in, to run ESM directly with my node@14.

Curiously, it seems to have a _lower_ performance, at around 11.5k rps. You can try it on your own: https://github.com/JaneJeon/validator-benchmark/tree/esm (you need to `npm i` for the `benchmarkify/` and `yup/` folders). 

[...]

PR #1022, opened 2020-08-28
perf: reduce function calls for shallower stacks

jquense:
starts to address #1017 bringing the benchmark for sync and async within striking distance of joi at least


src/util/createValidation.js
```diff
@@ -1,94 +1,41 @@
...
-      new ValidationError(formatError(message, params), value, path, type),
-      { params },
-    );
-  };
-}
-
-export default function createValidation(options) {
-  let { name, message, test, params } = options;
 
+export default function createValidation(config) {
```
jquense, line 5: most wins due to simplifying this call by reducing indirection and inlining helpers

src/util/runTests.js
```diff
@@ -0,0 +1,49 @@
+import ValidationError from '../ValidationError';
+import { once } from './async';
+
+export default function runTests(options, cb) {
```
jquense, line 4: more big wins here, by making this function less clear with the loop but much shallower

JaneJeon:
This is excellent! I'm getting 96k req/s with the synchronous implementation (vs. joi's 157k) - well within striking range of joi (50%), _and_ fast enough to not be the bottleneck for either express or fastify!

I'd love to see further work on this, and I'm willing to even contribute performance regression tests if necessary. Thanks for doing this!

jquense:
> and I'm willing to even contribute performance regression tests if necessary

That would be excellent and appreciated, having some benchmarks in the repo would be really helpful for tracking this stuff longer term. Doesn't need to be fancy just some numbers for yup to compare new versions against

JaneJeon:
will get on it after work!

PR #1023, opened 2020-08-28
Mark & resolve stale issues automatically

JaneJeon:
So one thing that was annoying me a bit was the number of issues/PRs that were essentially "dead" (i.e. no response from the original author) or resolved but _not_ closed. I think adding stalebot (and reducing the number of those stale issues) will help to clarify what issues actually _are_ active.

The only thing you need to do after merging this PR is to go to https://github.com/apps/stale and install stalebot on this repo.

Happy coding!

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 7
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 3
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+# Label to use when marking an issue as stale
+staleLabel: wontfix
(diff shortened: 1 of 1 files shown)


JaneJeon:
P.S. the performance PR is still coming, I just created this quickly during lunch break

Amoodaa:
@JaneJeon the performance PR never came rip,
im just kidding, thanks for that issue and that rigorous profiling !

JaneJeon:
@Amoodaa oh shoot, I completely forgot about this... if you’d like, you could pick up the mantle? https://github.com/JaneJeon/validator-benchmark/blob/master/suites/simple.js (*wink wink*)

Amoodaa:
@JaneJeon I would, ill be looking at it in the weekend then, first time doing that though

Amoodaa:
@JaneJeon can you tell me what should be done? reference to links is much appreciated :)

PR #1024, opened 2020-08-29
feat: use class syntax and static factories

jquense:


lucasff:
Why was this never merged?

PR #1029, opened 2020-09-04
fix: IE11 clone()

TheHalcyonSavant:
Moving `_whitelist` and `_blacklist` initializations inside `proto` fixes: `createSetIterator called on incompatible receiver` error in IE11

### DIFF
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -70,7 +70,6 @@ export default function SchemaType(options = {}) {
   this._conditions = [];
   this._options = { abortEarly: true, recursive: true };
   this._exclusive = Object.create(null);
-  
   this._whitelist = new RefSet();
   this._blacklist = new RefSet();
 
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -70,6 +70,7 @@ export default function SchemaType(options = {}) {
   this._conditions = [];
   this._options = { abortEarly: true, recursive: true };
   this._exclusive = Object.create(null);
+
   this._whitelist = new RefSet();
   this._blacklist = new RefSet();
 
@@ -97,7 +98,7 @@ const proto = (SchemaType.prototype = {
 
     // if the nested value is a schema we can skip cloning, since
     // they are already immutable
-    return cloneDeepWith(this, (value) => {
+    return cloneDeepWith(this, (value, key) => {
       if (isSchema(value) && value !== this) return value;
 
       // fix for ie11 when cloning Set and Map
--- a/src/mixed.js
+++ b/src/mixed.js
@@ -71,9 +71,6 @@ export default function SchemaType(options = {}) {
   this._options = { abortEarly: true, recursive: true };
   this._exclusive = Object.create(null);
 
-  this._whitelist = new RefSet();
-  this._blacklist = new RefSet();
-
   this.tests = [];
   this.transforms = [];
 
@@ -92,6 +89,9 @@ const proto = (SchemaType.prototype = {
   __isYupSchema__: true,
 
   constructor: SchemaType,
+  
+  _whitelist: new RefSet(),
+  _blacklist: new RefSet(),
 
   clone() {
     if (this._mutate) return this;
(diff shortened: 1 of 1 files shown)


jquense:
Not sure why CI didn't run here, but this likely breaks a lot of things...I don't think this is fixing the issue correctly, but it's hard to understand what the IE error even is here.

src/mixed.js
```diff
@@ -92,6 +89,9 @@ const proto = (SchemaType.prototype = {
   __isYupSchema__: true,
 
   constructor: SchemaType,
+  
+  _whitelist: new RefSet(),
+  _blacklist: new RefSet(),
```
jquense, line 94: this moves the properties to the prototype which is incorrect. They are instance properties and should not be shared between all instances of SchemaType, which is what would happen here.

TheHalcyonSavant:
To be able to replicate the IE11 failure, first prepare the build for es5 commonjs:
```js
const nodeResolve = require('rollup-plugin-node-resolve');
const commonjs = require('rollup-plugin-commonjs');
const babel = require('rollup-plugin-babel');
const filesize = require('rollup-plugin-filesize');

const base = {
  input: './src/index.js',
  plugins: [
    nodeResolve({
      main: true,
      browser: true
    }),
    commonjs(),
    babel({
      babelrc: false,
      presets: [[
        '@babel/preset-env', {
          targets: {
            ie: '11'
          }
        }
      ]],
    }),
  ],
};

module.exports = [
  {
    ...base,
    output: [
      {
        file: 'dist/yup.js',
        format: 'iife',
      },
    ],
    plugins: [...base.plugins, filesize()],
  },
];
```
Then, expose all yup functions to global window object from within src/index.js.
After running `./node_modules/.bin/rollup -c`, create the following html for testing:
```html
<html>
<head>
    <script src="some-default-es5-polyfills.js"></script>
    <script src="dist/yup.js"></script>
</head>
<body>
    <script>
        const object = window.yup.object;
        const string = window.yup.string;
        const number = window.yup.number;
        const date = window.yup.date;

        function testRun() {
            let s = string();
            try {
                s = s.required().concat(string()).concat(null);
            } catch (ex) {
                console.error(ex.stack);
            }
            const contactSchema = object().shape({
                name: s
            }).shape({
            age: number()
                .required()
                .positive()
                .integer(),
            email: string()
                .email(),
            website: string()
                .url(),
            createdOn: date()
                .default(function () { return new Date() })
            }).shape(null);

            contactSchema.isValid({
                // name: 'jimmy',
                age: 24,
                email: 'jdog@cool.biz'
            }).then(console.log);

            contactSchema.isValid({
                name: 'jimmy',
                age: 24,
                email: 'jdog@cool.biz'
            }).then(console.log);
        }

        testRun();
    </script>
</body>
</html>
```
Running this in IE11 produces breaks on `concat` in the try catch block. Further investigation will show that `this._whitelist.list.size` has `<Permission denied>`:
image

My PR fixes this. It doesn't break anything else as far as I know.

jquense:
This only fixes the issue because it avoids trying to clone the RefSets, but it creates a subtle and bad bug where one instance of these RefSets is shared between all schema globally. Judging from the error it looks like the actual bug is in the Concat method, but not sure where. And to be clear even with es6 classes this would still not work. The code as it's is functionally the same as an es6 class

TheHalcyonSavant:
Thanks for responding!

I ran my local tests badly, which misled me that everything was fine.
The bug seems very deep, not just with concat. I’ve seen it elsewhere with cloning and chaining of RefSet, but I can’t reproduce it for other cases. So far, I can reproduce it only for `string().required().concat(string())`, therefore I will remove this usage and continue.

TheHalcyonSavant:
Also, `boolean().oneOf([true])` fails in IE11

TheHalcyonSavant:
For some reason, the RefSets aren't cloned during the SchemaType.clone and this causes all of the IE11 bugs.
My guess is that other browsers don’t lock the inherited `size` property, and that’s why they can successfully clone Sets and Maps. Or maybe the bug could be inside lodash `cloneDeepWith` or in the polyfills, but I can’t debug any further as I have a solution.

This new commit doesn’t break any of the 536 tests, would you accept it?

jquense (approved):

PR #1050, opened 2020-09-29
Typescript rewrite

jquense:
rewritten in typescript, the interesting bit here is the new approach to typing the input and output of yup schema. I was thinking this uncomfortably pushes to the TS type system usage into places it wasn't intended but with TS 4.1 ppl are writing whole sql parsers in it so maybe this isn't as far fetched as i thought :P 

Having this finally written in TS has made it a lot safer to remove extra code, and helpfully forcing more constrained interfaces on all Schema. In particular new Schema can store extra instance metadata in `spec` which we can automatically clone and propagate when schema change. This means that the base `clone` and `concat` methods can do much less deep cloning, b/c schema instances are no longer functioning as fly-by-night records.

**Typing approach:**

Yup schema basically track **two** output types, the output of `cast`, e.g. the transformed input. and the validated type, which is the cast return type _plus_ additional knowledge about presence. Take a look at the types test to see why these are different and why it matters. The excellent TD types track nullability and presence via augmenting the output type as they are set, e.g. by removing `null` from `string | null` as the output of `nullable(false)`. Here i'm trying an alternative which instead tracks the config state of the schema, which is used to build final output types. So instead of `nullable()` making the schema type `string` -> `string | null`, you get `StringSchema<string, 'nullable'>, this allows us to track more specific details like whether `nullable` is explicitly set, or the default false. It's all kinda wonky, but does seem to work. I need to take stock again of whether it's worth it over just changing the type.

**Typing goals:**

- `cast` and `validate` both return as close as correct types as possible, accounting for defaults, and not requiring strict() to be accurate.
- Inferred typed should constrain output of these yup methods _not_ the input. 
- Produce practically helpful types over strict correctness. generally these are the same thing but Yup's API is really not written in a "type safe" way, and it's not intended to be. Yup is not a type system, it's a DSL for writing assertions and transforms on anything you want. Which makes is challenging to produce fully correct types, partly b/c it can produce unsound types. 

**TODO:**

- incomplete abstract base Schema interface, there is an incorrect mix of things expecting MixedSchema when they should be expecting Schema
- `concat` is still wrong and not sure how much time i want to spend trying to get it working. trying to type it has surfaced a lot of inconsistencies between how we merge nullability and presence, special casing when these are "unset" vs explicilty set to false/true
- Utils aren't typed correctly yet
- I'm sure there is weirdness around arrays


jquense:
cc @itajaja @iansan5653 @mauricedb for input if ya'll have time

jensbodal:
Something to consider is that you can do this incrementally. 

* add support for typescript and modify build system to support the hybrid setup
* start from the leafs and convert them over to typescript 
* keep iterating from there until complete

This will have the added benefit of 
* separating out the addition of typescript from the full on conversation
* isolating any breaking changes to a smaller set of changes
* making each piece that might need to be refactored easier to reason with

Maybe you’re already partially doing this, it just seems way more complicated to submit a CR for 50+ files and a somewhat complete rewrite when it could be done piece by piece instead.

Bartmr:
Are there any plans on when will this be finished? Is there any way we can help you? I have some expertise on Typescript. My current company has some isomorphic apps and keeping the types linked between stacks has given me some experience, specially types that transform themselves (like converting all Date properties in a deep object into strings, to mirror the effects of serialization) .

jquense:
You can help out by trying out the new code here or reviewing the type defs. Ideally i'd like to get a sense of where the types fail

Bartmr:
I was able to fix some typing issues in `prependDeep` and also enabled Javascript in `tsconfig` so the compiler allows JS imports.

#1060

dko-slapdash:
JFYI sharing my opinion on the TS rewrite (no offense, just what's on my mind - hoping it would resonate). I think there are drawbacks in changing the main argument's type from being a pure TS interface type to being an ObjectShape. Mainly because it produces a way less nice TS compiler's hinds in e.g. IntelliSense and in type error messages:

image

I think it can be what drains the library's popularity long term and even kills it.

E.g. in the code:

```
const schema = object({ aaa: string().required() });
typeof schema
```

in the old typing from DefinitelyTyped, the schema's type was something like `Schema<{ aaa: string }>` (which is easily readable), and now it's `ObjectShape<...very-long-chunk...>`. For large schemas, it makes the output really long.

Also, in the previous typing, one could do the following thing: `object<{ aaa?: string }>({ aaa: string() })` and thus have "optional undefineds" for free. It's not possible in the new code since ObjectShape resists to changes of the TS underlying type.

dko-slapdash:
Another downside is the size of the generated *.d.ts files when a schema is exported. This affects tsc compilation speed a lot.

E.g. for the following simple schema:

image

this is what tsc generates in *.d.ts file when `declaration=true` is turned on:

image

Notice how large and repetitive it is.

Bartmr:
@dko-slapdash  If you have incremental builds, it will only need to process that once. And even with that, you can disable such .d.ts generation if you're not doing a library. Web apps and servers only need code + source maps.

dko-slapdash:
The whole point is that we need these d.ts files while developing (monorepo setup with multiple packages), because we need typing info in IntelliSense. My comment was about the inefficiency of the approach when the generic argument type is passed as ObjectShape and not as a plain TS type (as in the DT version prior to the rewrite, and also as in superstruct). This is kinda pattern: the simpler the generic types are, the faster is the build, and the less memory tsc uses.

PR #1061, opened 2020-10-06
Add method to allow empty array when using along .required()

ejose19:
Proposed PR adds a method called ```allowEmpty``` that allows empty arrays to be valid when ```required``` is also set.

Example usage:

```ts
const schema = array().required().allowEmpty();
schema.isValidSync([]); // true
```

Also, added a missing test dependency and fixed object tests that had 2 typos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,8 @@
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-size-snapshot": "^0.12.0",
     "sinon": "^9.0.3",
-    "sinon-chai": "^3.5.0"
+    "sinon-chai": "^3.5.0",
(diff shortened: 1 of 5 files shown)


jquense:
you can already use `defined()` for what you want i believe

ejose19:
@jquense defined allows null as valid, and for some that's not a desired valid value.

jquense:
> defined() allows null as valid 

that's a bug tho. defined should not set `nullable()`

ejose19:
if that behaviour will be removed, then this can be dropped. However should the fixes to tests be merged? That will allow new contributors to ```yarn && yarn test``` without errors.

PR #1089, opened 2020-10-29
docs: Clarify return value of mixed.test

vapurrmaid:
## Context

While it may seem obvious to many, I've previously found myself searching the docs more than once for
whether or not I wanted to return `true` or `false` in `mixed.test` validators. This is a minor change to
the README that clarifies the return value.

## Additional Details

Cleaned up some long lines and capitalized the beginning of each paragraph

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,14 +692,14 @@ In order to allow asynchronous custom validations _all_ (or no) tests are run as
 A consequence of this is that test execution order cannot be guaranteed.
 
 All tests must provide a `name`, an error `message` and a validation function that must return
-`true` or `false` or a `ValidationError`. To make a test async return a promise that resolves `true`
-or `false` or a `ValidationError`.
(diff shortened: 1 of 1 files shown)


jquense (approved):
thanks!

README.md
```diff
@@ -720,7 +720,10 @@ await schema.isValid('jimmy'); // => true
 await schema.isValid('john'); // => false
 ```
 
-test functions are called with a special context, or `this` value, that exposes some useful metadata and functions. Older versions just expose the `this` context using `function ()`, not arrow-func, but now it's exposed too as a second argument of the test functions. It's allow you decide which approach you prefer. 
+Test functions are called with a special context, or `this` value, that exposes some useful metadata
+and functions. Older versions just expose the `this` context using `function ()`, not arrow-func,
```
mattclough1, line 724: Is this true for the current release?

PR #1090, opened 2020-10-29
fix README documentation of mixed.test

probablyup:
According to the TS types and real-world usage, context seems to be the first argument, not the second argument.

Based on yup@0.29.3


jquense:
that would be a bug if this is the case. I don't see this tho...

probablyup:
After some more work on this, you're right. I think the types are out of date

PR #1112, opened 2020-11-16
Change uuid's regexp

RickyFeria:
Change in uuid regexp to more exactly meet the specs defined in RFC 4122. The regexp matches the string with the following the format: xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx where x is one of [0-9, a-f] M is one of [1-5], and N is [8, 9, a, or b] and 00000000-0000-0000-0000-000000000000 is the zero value.


jquense:
thanks!

PR #1134, opened 2020-12-01
Ts simpler 2

jquense:
Ok this is what i've settled on, it's most similar to the DT types, and fairly verbose but not doing any weird typing things and most of the verbosity is in interfaces that will be compiled away.


README.md
```diff
@@ -1326,84 +1356,53 @@ const personSchema = yup.object({
...
 You can derive the TypeScript type as follows:
 
-```TypeScript
-type Person = yup.InferType<typeof personSchema>;
-```
+```ts
+import type { Asserts, TypeOf } from 'yup'
 
-Which is equivalent to the following TypeScript type alias:
+type parsed: Typeof<typeof personSchema> = personSchema.cast(json);
```
jquense, line 1379:
```suggestion
const parsed: Typeof<typeof personSchema> = personSchema.cast(json);
```

src/string.ts
```diff
@@ -0,0 +1,279 @@
...
+// eslint-disable-next-line
+let rUUID = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+
+let isTrimmed = (value: Maybe<string>) =>
```
joaopaulobdac, line 15: Is there a reason for this function being up here in the file instead of in the `trim` test? It's not being called from anywhere else :)

lookfirst:
I ❤️ exclusiveTests. Thank you for making this public.

jquense:
it's not public!

lookfirst:
Ok, sshhh... it isn't public. ;-)

gotexis (commented):
looks like ie11 is broken in this pr

PR #1148, opened 2020-12-04
Add additional functions to Lazy class

chasecaleb:
Add missing functions to Lazy:

    isValid
    isValidSync
    isType

These are all validation functions that have a value parameter, so it makes sense for Lazy to wrap them.

### DIFF
--- a/src/Lazy.ts
+++ b/src/Lazy.ts
@@ -89,9 +89,6 @@ class Lazy<T extends AnySchema, TContext = ContextOf<T>>
   isValidSync(value: any, options?: ValidateOptions<TContext>) {
     return this._resolve(value, options).isValidSync(value, options);
   }
-  isType(value: any) {
-    return this._resolve(value).isType(value);
-  }
 }
 
 export default Lazy;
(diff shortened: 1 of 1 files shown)


src/Lazy.ts
```diff
@@ -82,6 +82,16 @@ class Lazy<T extends AnySchema, TContext = ContextOf<T>>
...
+
+  isValid(value: any, options?: ValidateOptions<TContext>) {
+    return this._resolve(value, options).isValid(value, options);
+  }
+  isValidSync(value: any, options?: ValidateOptions<TContext>) {
+    return this._resolve(value, options).isValidSync(value, options);
+  }
+  isType(value: any) {
+    return this._resolve(value).isType(value);
+  }
```
jquense, line 94: hmm without the resolve options this will likely return the wrong thing, or be incorrect, but adding options here makes the method signature different than the `schema.isType`...maybe better not to add this one?

chasecaleb: Hmm, I thought I was using it successfully in my own project but apparently not, so I'm not sure. I went ahead and removed it.

jquense (approved):

jquense:
thanks!

chasecaleb:
Thanks for merging! Do you think you'll be able to make a release of this on npm soon? That's the last thing preventing me from switching back to using your version.

PR #1166, opened 2020-12-10
Make StringSchema.matches options optional

youngbob:
The second parameter to matches() should be optional.

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -105,7 +105,7 @@ export default class StringSchema<
     });
   }
 
-  matches(regex: RegExp, options: MatchOptions | MatchOptions['message']) {
+  matches(regex: RegExp, options?: MatchOptions | MatchOptions['message']) {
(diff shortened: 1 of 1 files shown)


src/string.ts
```diff
@@ -105,26 +105,24 @@ export default class StringSchema<
     });
   }
 
-  matches(regex: RegExp, options: MatchOptions | MatchOptions['message']) {
+  matches(regex: RegExp, options: MatchOptions | MatchOptions['message'] = locale.matches) {
```
jquense, line 108: this is intentionally not being defaulted up here because it can mess with the overload check below, e.g. messages can be objects. I'd just mark this as optional instead

youngbob: Ok. Reverted and just marked it optional.

jquense (approved):

jquense:
thanks!

PR #1169, opened 2020-12-10
Fix array handling in SchemaOf type

chasecaleb:
1. Move array before object so that inference works correctly
1. Include `Lazy` as part of `ArraySchema`'s generic, to match the class definition

Here's a quick snippet that illustrates why 1 is necessary:
```typescript
export const array: Array<string> = ["abc"];
// Valid - an Array is a Record (note: AnyObject is a type alias for Record<string, any>)
export const record: Record<string, any> = array;
// Doesn't compile
export const anotherArray: Array<string> = record;
```

By the way, I noticed that eslint isn't checking the `$ExpectType` comments in `test/type.ts`... I investigated it for a little bit, but ran into more problems so I gave up. You need to add `--ext=.ts,.js` when running eslint to make sure it checks the ts file, but then eslint started to give parser errors and I lost motivation. That's my excuse for not adding tests, sorry about that.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,7 +6,7 @@ import DateSchema, { create as dateCreate } from './date';
 import ObjectSchema, { AnyObject, create as objectCreate } from './object';
 import ArraySchema, { create as arrayCreate } from './array';
 import { create as refCreate } from './Reference';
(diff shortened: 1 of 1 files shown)


jquense:
huh TIL. thanks!

Issue #1155, opened 2020-12-07
"This expression is not callable." when trying to use the promise returned by .validate() of a generic schema in v0.32


justincy:
**Describe the bug**

In v0.32, I get a tslint error saying `This expression is not callable.` when calling `.then()` or `.catch()` on the return value of `.validate()` on a generic schema.

```
This expression is not callable.
  Each member of the union type '(<TResult1 = AssertsShape<{ [k in keyof T]: SchemaOf<T[k]>; }> | Extract<TypeOfShape<{ [k in keyof T]: SchemaOf<T[k]>; }>, null | undefined>, TResult2 = never>(onfulfilled?: ((value: AssertsShape<...> | Extract<...>) => TResult1 | PromiseLike<...>) | ... 1 more ... | undefined, onrejected?: ((reason: any) => TResult...' has signatures, but none of those signatures are compatible with each other.
```

**To Reproduce**

https://codesandbox.io/s/yup-v032-react-final-form-typescript-ogkmq?file=/src/components/SimpleForm.tsx

**Expected behavior**

In the same codesandbox as linked to above, if you look at the [App](https://codesandbox.io/s/yup-v032-react-final-form-typescript-ogkmq?file=/src/components/App.tsx) component you'll see similar code that works because it's not using a generic.

**Additional context**

The return type of `.validate()` in the case of the generic is quite different than the non-generic case.

*Generic*

```ts
BaseSchema<TCast = any, TContext = Record<string, any>, TOutput = any>.validate(value: any, options?: ValidateOptions<Record<string, any>> | undefined): Promise<AssertsShape<{ [k in keyof T]: Yup.SchemaOf<T[k]>; }> | Extract<...>> | Promise<...> | Promise<...>
```

*Specific*

```ts
BaseSchema<TypeOfShape<{ email: BaseSchema<Maybe<string>, Record<string, any>, string>; password: BaseSchema<Maybe<string>, Record<string, any>, string>; }>, Record<...>, AssertsShape<...>>.validate(value: any, options?: ValidateOptions<Record<string, any>> | undefined): Promise<AssertsShape<{
    email: Yup.BaseSchema<Maybe<string>, Record<string, any>, string>;
    password: Yup.BaseSchema<...>;
}>>
```

Seems relevant: https://stackoverflow.com/a/49511416/8786413

jquense:
[...]

PR #1170, opened 2020-12-10
Fix ObjectSchemaOf type definition

chasecaleb:
Follow up to #1169, because in my haste to rebase I overlooked adding `Lazy` to the `ArraySchema` generic in `ObjectSchemaOf`. I also simplified the `ObjectSchemaOf` definition since it doesn't need to repeat the `SchemaOf` definition. I also fixed my incorrect formatting from #1169.

I've tested this against my own project via `yarn link` so I know it at least works for my use cases.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,19 +40,19 @@ function addMethod(schemaType: any, name: string, fn: any) {
 type ObjectSchemaOf<T extends AnyObject> = ObjectSchema<
   {
     [k in keyof T]-?: T[k] extends Array<infer E>
-      ? ArraySchema<SchemaOf<E>>
+      ? ArraySchema<SchemaOf<E> | Lazy<SchemaOf<E>>>
       : T[k] extends AnyObject
       ? // we can't use  ObjectSchema<{ []: SchemaOf<T[k]> }> b/c TS produces a union of two schema
-        ObjectSchemaOf<T[k]>
+        ObjectSchemaOf<T[k]> | ObjectSchemaOf<Lazy<T[k]>>
       : BaseSchema<Maybe<T[k]>, AnyObject, T[k]>;
   }
 >;
 
 type SchemaOf<T> = T extends Array<infer E>
-    ? ArraySchema<SchemaOf<E> | Lazy<SchemaOf<E>>>
-    : T extends AnyObject
-    ? ObjectSchemaOf<T>
-    : BaseSchema<Maybe<T>, AnyObject, T>;
+  ? ArraySchema<SchemaOf<E> | Lazy<SchemaOf<E>>>
+  : T extends AnyObject
+  ? ObjectSchemaOf<T>
+  : BaseSchema<Maybe<T>, AnyObject, T>;
 
 export type AnyObjectSchema = ObjectSchema<any, any, any, any>;
 


jquense:
I actually don't think you want to add the Lazy here. What is the use case?

chasecaleb:
Something like this, where the snippet below is used for a field in a `Yup.object()` validator

```typescript
Yup.array(
        Yup.lazy<YupSchema<ModuleData>>(element => {
            // ... dynamic logic here
        })
    );
```

chasecaleb:
BTW, #1169 already added Lazy to the ArraySchema generic within `SchemaOf`. This PR just makes it consistent within `ObjectSchemaOf` as well.

jquense:
yeah i see. hmm I am a bit concerned that this will break other common cases like https://github.com/jquense/yup/issues/1155 

It probably makes sense to have a common interface shared between lazy and schema (there is already one simple one) but unsure if that would break other stuff...

chasecaleb:
I don't think it breaks that use case, does it? Changing the type definition to add `Lazy` to the top level like `SchemaOf<T> = Lazy<T> | ...the rest` would cause issues, but that's not what I'm proposing here. The only part I'm adding `Lazy` to is inside `ArraySchema<SchemaOf<E> | Lazy<SchemaOf<E>>>`, not at the top level. So you would still be calling `validate()` on the `ArraySchema` type, which the inner generic part shouldn't affect.

jquense:
yeah that's true for this specific change, but technically it's an incomplete change, since object properties also accept Lazy

so you fix your example above but this still doesn't work i think?
```
type Person {
   name: string
}

const s: SchemaOf<Person> = object({ name: lazy(() => string().defined() })
```

src/index.ts
```diff
@@ -39,20 +39,15 @@ function addMethod(schemaType: any, name: string, fn: any) {
 
 type ObjectSchemaOf<T extends AnyObject> = ObjectSchema<
   {
-    [k in keyof T]-?: T[k] extends Array<infer E>
-      ? ArraySchema<SchemaOf<E>>
-      : T[k] extends AnyObject
-      ? // we can't use  ObjectSchema<{ []: SchemaOf<T[k]> }> b/c TS produces a union of two schema
-        ObjectSchemaOf<T[k]>
-      : BaseSchema<Maybe<T[k]>, AnyObject, T[k]>;
+    [k in keyof T]-?: SchemaOf<T[k]>;
```
chasecaleb, line 42: Regarding the the nested-lazy-inside-object concern, how about this? That doesn't cause any compile issues in my codebase, so I think it's valid.
```suggestion
    [k in keyof T]-?: SchemaOf<T[k]> | Lazy<SchemaOf<T[k]>>;
```

jquense: we can try it!

chasecaleb: And doing it like this still avoids putting `Lazy` in the top level of `SchemaOf` so it's still safe in that regard too.

chasecaleb:
Okay, I just committed the `ObjectSchemaOf` addition. How's it look now?

src/index.ts
```diff
@@ -39,20 +39,15 @@ function addMethod(schemaType: any, name: string, fn: any) {
 
 type ObjectSchemaOf<T extends AnyObject> = ObjectSchema<
   {
-    [k in keyof T]-?: T[k] extends Array<infer E>
-      ? ArraySchema<SchemaOf<E>>
-      : T[k] extends AnyObject
-      ? // we can't use  ObjectSchema<{ []: SchemaOf<T[k]> }> b/c TS produces a union of two schema
-        ObjectSchemaOf<T[k]>
-      : BaseSchema<Maybe<T[k]>, AnyObject, T[k]>;
+    [k in keyof T]-?: SchemaOf<T[k]> | Lazy<SchemaOf<T[k]>>;
```
jquense, line 42: see the old comment here. We can't use `SchemaOf<T[k]>` because of how TS applies the types over the conditional in SchemaOf. Basically if the input type is something like `string | undefined`, you get `SchemaOf<string> | SchemaOf<undefined>`, not `SchemaOf<string | undefined>`, The behavior is different for passing the generic to a concrete type like `BaseSchema`. Just means some duplication

chasecaleb: Okay... I'm really confused about what's going on here. Either I'm doing something wrong, or it seems like that comment is backwards. Can you look at the minimal example I made in this commit? https://github.com/jquense/yup/pull/1170/commits/8bf7689021379cc2ad47f3c34f705dd556b8889f

chasecaleb: I somewhat suspect the issue is with my example code (`FormData`/`buildNestedValidator`), not the Yup type definitions... but I can't figure out the right way, either. I'm not sure how to translate my recursive `FormData` to a corresponding `SchemaOf` type.

chasecaleb: Ah ha, I think I figured out what's going on. I'll get back to you soon with this sorted out. Recursive types are... fun.

chasecaleb: Okay, I added a commit that adds lazy to field values for object schemas.

And in case you wanted an explanation on how I got so confused:
```typescript
interface FormData {
  [K: string]: string | FormData;
}
type FormDataSchema = ObjectSchema<{
  [K: string]: SchemaOf<string> | FormDataSchema;
}>;
export const buildNestedValidator = (
  path: string[],
  validator: SchemaOf<string> | FormDataSchema,
): FormDataSchema => {
  const [head, ...rest] = path;
  const result: SchemaOf<string> | FormDataSchema =
    rest.length === 0 ? validator : buildNestedValidator(rest, validator);
  return objectCreate({ [head]: result });
};
const foo: FormData = { foo: '', bar: { baz: '' } };
const val = buildNestedValidator(['bar', 'baz'], stringCreate().required());
val.validate(foo);
```

I was trying to use `SchemaOf<FormData>` originally, but that wasn't working. The solution was to create the `FormDataSchema` type and use it instead. I think that's because of how the union gets distributed - e.g. `SchemaOf<string | string[]>` vs `SchemaOf<string> | SchemaOf<string[]`. Something along those lines - my brain hurts at the moment.

Anyways, that's all irrelevant to the actual change in this PR and is just context on how I got so tripped up testing against my own code base. I went ahead and rebased to squash my mess - **this is ready for you to review and potentially merge now**.

chasecaleb:
Hi, any thoughts on the current state of this MR?

jquense:
will get back to you next week, on vacation!

chasecaleb:
@jquense quick reminder about this PR. Hope you had a good vacation, too!

chasecaleb:
Thanks for working with me on this PR and merging it!

Do you have plans to make a new release with this change? It would help me out when you get a chance so that I can get rid of some temporary workaround code, but it isn't urgent.

PR #1177, opened 2020-12-12
Do not initialize spec values with undefined

### LINKED ISSUE #1160: baseSchema.concat(otherSchema) overwrites baseSchema properties when otherSchema properties are undefined
**Describe the bug**
Last_Edit: It is because of this code https://github.com/jquense/yup/blob/master/src/schema.ts#L231-L233 , `combined.spec` has every property `undefined`, it would work if `combined.spec` properties were non-existent instead of `undefined`:

`const mergedSpec = { ...base.spec, ...combined.spec };`

First_Edit: Looking a bit deeper, i am now not sure if it is caused by mixed.label(), but since the only thing we change is the yup version and this unexpected behavior occurs, i will keep this issue as is, hopefully someone experienced in yup can investigate or realize the reason why this happens.

Calling mixed.label("Custom label") does work on version 0.31.0 but not on 0.32.5+.
Example screenshots from v.0.31.0 and v.0.32.5+
v.0.31.0:

before_update

v.0.32.5+:

after_update

**To Reproduce**
Quite complex form logic react and formik, will try to extract the logic and edit here.

**Expected behavior**
I would expect the previous behavior of setting a label to work on error messages.

 - Browser [chrome]
 - Version [86]

**Additional context**
Using with react + formik

cemalettin-simsoft:
Otherwise spreading spec values like previously would overwrite the base schema spec values with undefined.
Closes #1160.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -137,8 +137,6 @@ export default abstract class BaseSchema<
       strict: false,
       abortEarly: true,
       recursive: true,
-      label: undefined,
-      meta: undefined,
       nullable: false,
       presence: 'optional',
       ...options?.spec,
--- a/test/mixed.js
+++ b/test/mixed.js
@@ -645,6 +645,19 @@ describe('Mixed Types ', () => {
     }.should.throw(TypeError));
   });
 
+  it('concat should not overwrite label and meta with undefined', function () {
+    const testLabel = "Test Label"
+    const testMeta = {
+      testField: "test field"
+    }
+    let baseSchema = mixed().label(testLabel).meta(testMeta)
+    const otherSchema = mixed()
+
+    baseSchema = baseSchema.concat(otherSchema)
+    expect(baseSchema.spec.label).to.equal(testLabel)
+    expect(baseSchema.spec.meta.testField).to.equal(testMeta.testField)
+  })
+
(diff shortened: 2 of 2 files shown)


jquense:
thanks for the PR, i'd need some context for why this behavior should be changed and tests as well

jquense:
oops sorry i didn't see the linked issue 👍

src/schema.ts
```diff
@@ -229,7 +229,14 @@ export default abstract class BaseSchema<
     let base = this;
     let combined = schema.clone();
 
-    const mergedSpec = { ...base.spec, ...combined.spec };
+    const baseSpec = { ...base.spec };
+    // only overwrite with spec values which are not undefined
+    const mergedSpec = Object.keys(combined.spec).reduce((accum, key) => {
```
jquense, line 234: i think it probably simpler to not define undefined properties in the spec instead and keep the object spread

cemalettin-simsoft: Yes, that was the other option, i wasn't really sure about editing that part, it may depend on other stuff as well, not too familiar with yup.

jquense (approved):

jquense:
thanks!

PR #1178, opened 2020-12-12
Readonly array type for function arguments

gcangussu:
### Why

Functions that doesn't require a mutable array must be declared with readonly arrays. Mutable arrays extends readonly arrays, so we can provide a mutable array to a function accepting readonly arrays. But, if we declare a function accepting a mutable array, we cannot provide a readonly array.

Example
```ts
const MY_ENUM_VALS = ['one', 'two', 'three'] as const;

// error: 'readonly ["one", "two", "three"]' is not assignable to the mutable type 'string[]'
yup.string().oneOf(MY_ENUM_VALS);
```

### What & How
Changed function declarations that accepted arrays to accept readonly arrays. I've checked the implementation, we are not mutating arrays received as function arguments.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -625,13 +625,13 @@ export default abstract class BaseSchema<
   }
 
   when(options: ConditionOptions<this>): this;
-  when(keys: string | string[], options: ConditionOptions<this>): this;
+  when(keys: string | readonly string[], options: ConditionOptions<this>): this;
   when(
-    keys: string | string[] | ConditionOptions<this>,
+    keys: string | readonly string[] | ConditionOptions<this>,
     options?: ConditionOptions<this>,
   ) {
     if (!Array.isArray(keys) && typeof keys !== 'string') {
-      options = keys;
+      options = keys as ConditionOptions<this>;
       keys = '.';
     }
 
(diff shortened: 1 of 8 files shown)


src/schema.ts
```diff
@@ -625,13 +625,13 @@ export default abstract class BaseSchema<
...
-  when(keys: string | string[], options: ConditionOptions<this>): this;
+  when(keys: string | readonly string[], options: ConditionOptions<this>): this;
   when(
-    keys: string | string[] | ConditionOptions<this>,
+    keys: string | readonly string[] | ConditionOptions<this>,
     options?: ConditionOptions<this>,
   ) {
     if (!Array.isArray(keys) && typeof keys !== 'string') {
-      options = keys;
+      options = keys as ConditionOptions<this>;
```
gcangussu, line 634: typescript 4.1.2 is able to infer the correct type of `keys`, given the type guards. Let me know if you prefer to update the typescript version rather than this type assertion.

xiaoyu-tamu:
Hi @jquense, I'm experiencing the same issue.   what the status of this PR?

xiaoyu-tamu:
I noticed that 1.0.0.alpha fixed this issues, is that branch gonna replace 0.30 in the future?

gcangussu:
@jquense Let me know if you have interest in this so I can solve the conflicts.

PR #1187, opened 2020-12-16
upgraded lodash-es min version from 4.17.11 to 4.17.15

WCDave:
upgraded lodash-es min version from 4.17.11 to 4.17.15 due to security vulnerabilty found by Checkmarx

ran npm run test...all passed


jquense:
thanks!

PR #1194, opened 2020-12-22
reach and getIn make last 2 arguments optional

### LINKED ISSUE #1180: react function last options
last two options were optional. but they are not. typescript create error. 

https://github.com/jquense/yup/blob/e785e1a4ddf1b7fb2d1ef48cb258d8e21a775446/src/util/reach.ts#L53

mattywong:
makes the last 2 arguments of the reach utility function (and getIn) optional 

closes #1180


jquense:
Thanks!

Issue #1116, opened 2020-11-19
validate and validateAt have different inputs for the when validators


belgac:
Hi,

I am using both validateAt and validate methods in my codebase, but when i use validateAt the dependency array from the when validators is not casted.

How should i write my when function if i can't rely on casting?

here is a runnable test demonstrating the problem:
https://codesandbox.io/s/magical-browser-i3210

I already tried to create a custom method to cast the values when you are using at, relying on a context trick but it only works for the first level of depth of the object and when doesn't give the path in this
```js
const getCastedDeps = (context, deps, originalDeps) => {
  const { validator, rootValue, at } = context || {};
  return at
    ? [
        ...deps.map((dep, index) => {
          if (dep.startsWith("$")) {
            return originalDeps[index];
          }
          try {
            return validator.validateSyncAt(dep, rootValue, { context });
          } catch (err) {
            return null;
          }
        }),
        originalDeps[originalDeps.length - 2],
        originalDeps[originalDeps.length - 1]
      ]
    : originalDeps;
};

addMethod(mixed, 'whenAlwaysCast', function(deps, fn) {
  return this.when(['$', ...deps], (context, ...originalDeps) => {
    const castedDeps = getCastedDeps(context, deps, originalDeps);
    return fn(...castedDeps);
  })
})
```

belgac:
ok i made it work using the path in the last argument received from '.when', wich is the schema, it seems a bit hacky.

if anyone has a better solution, i welcome it.

it's still kind of unexpected to me and i think we should put a warning about this in the docs. What do you think?

https://codesandbox.io/s/infallible-tu-epoqz?file=/src/index.test.js

PR #1197, opened 2020-12-24
doc(validateAt): add sibling casting warning with validateAt

belgac:
the validate and validateAt method differ in the fact that validateAt doesn't cast the siblings dependencies of when at the nested path (see #1116 ) I think this should be warned for in the docs to not surprise the user of yup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,6 +460,8 @@ but uses the resulting schema as the subject for validation.
 
 > Note! The `value` here is the _root_ value relative to the starting schema, not the value at the nested path.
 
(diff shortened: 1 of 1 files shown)


jquense:
I hmm, this may be something we should fix. I need to think through if that's feasible tho

belgac:
Hello! where are we on this issue?

jquense:
i've not thought much about it. happy to take a PR

belgac:
ok so it should be implemented and not just warned in the docs, i will try to find some  time in my schedule, thank you ;)

Issue #1162, opened 2020-12-10
Can't export MixedSchema when declaration=true set in tsconfig


Gohsato:
I'm trying to make a Typescript package with a yup schema.  
My plan is to import this package in a couple projects and use the same validator.  

I'm having some difficulty exporting the schema.   
What I'm trying to do is export a mixed schema like so. But I'm getting a compiler error.
```
export enumSchema = yup.mixed<MyEnum>.().oneOf(Object.values(MyEnum))
```

The following example summarizes the problem I'm having.  
```
// Set declaration:true in tsconfig.json
import { MixedSchema } from "yup";
export const a = new MixedSchema();
```
Which produces an error:

> Exported variable 'a' has or is using name 'MixedSchema' from external module "/sandbox/node_modules/yup/lib/mixed" 
> but cannot be named.ts(4023)

I've also tried a few variations without success.  
```
export const a = yup.mixed();
export const a: MixedSchema = yup.mixed();
```

How do I export 'a'?  
Appreciate the help!

Here's the sandbox: https://codesandbox.io/s/pedantic-ishizaka-f7th5?file=/src/index.ts

jquense:
i'm not really clear on what this error means...usually it's because types aren't exported by the package, but they are so unsure why this is happening...

Gohsato:
I just did some digging and it seems like these few lines may be the issue
https://github.com/jquense/yup/blob/master/src/mixed.ts#L38-L40
```
const Mixed: typeof MixedSchema = BaseSchema as any;
export default Mixed;
```
When Mixed is exported in this way only the class constructor is exported, not the class instance types.  (This doc explains it nicely https://www.typescriptlang.org/docs/handbook/classes.html#constructor-functions).    

One way to fix it would be to use namespace merging. 
```
export class MixedSchema extends BaseSchema{}
export interface MixedSchema extends BaseSchema{
// migrate types from 'declare class MixedSchema'
}
```
I have created a sample implementation that shows how it will work in the sandbox.  
https://codesandbox.io/s/falling-sun-c15fx

jquense:
hmm, the code is currently like this for a reason, Mixed, needs different types than Base, but i don't actually want them to be distinct classes. extending Mixed, should extend all other schema (it's weird)

Gohsato:
So if I understand correctly:   
Mixed should have the **exact** same implementation as Base. But Mixed has different type signatures for its functions.

jquense:
[...]

PR #1204, opened 2021-01-03
Export MixedSchema to fix ts with --declarations

janicduplessis:
Fixes ts error `Exported variable 'Schema' has or is using name 'MixedSchema' from external module "x/node_modules/yup/lib/mixed" but cannot be named.` with --declarations builds.

The MixedSchema class needs to be exported so it can be used in declaration files.

### DIFF
--- a/src/mixed.ts
+++ b/src/mixed.ts
@@ -4,7 +4,7 @@ import { AnyObject, Maybe, Optionals } from './types';
 import type { Defined } from './util/types';
 import BaseSchema from './schema';
 
-declare class MixedSchema<
+export declare class MixedSchema<
   TType = any,
   TContext = AnyObject,
   TOut = TType


Gohsato:
I think this pr would solve #1162!

MartinCerny-awin:
Can we please merge this PR?

jquense:
Haven't had the time to try this out, but hoping to make some this week

Shandur:
@jquense Looking forward to seeing it merged too 👍

dko-slapdash:
+1

It's a simple change consistent to what e.g. object.d.ts or number.d.ts have, and no risk of regressions @jquense - the button "Merge" can just be pressed without even testing.

jquense:
every TS change needs to be tested, unfortunately @dko-slapdash the smallest change can have large, or to picture effects.

janicduplessis:
Thanks @jquense !

pkolt:
@jquense Hi! When will the version with this patch be released?

PR #1208, opened 2021-01-05
Add generic to Reference.create()

florianeckerstorfer:
The `create()` method in `Reference.ts` was missing the generic, which caused a type error in the following example

```typescript
yup
  .object({
    availableFrom: yup.number(),
    availableTo: yup
      .number()
      .moreThan(yup.ref('availableFrom')),
  });
```

Error:
```
Argument of type 'Reference<unknown>' is not assignable to parameter of type 'number | Reference<number>'.
```

In this PR I add a generic to the `create()` method and type `Reference` and `ReferenceOptions` with it.

### DIFF
--- a/src/Reference.ts
+++ b/src/Reference.ts
@@ -10,8 +10,8 @@ export type ReferenceOptions<TValue = unknown> = {
   map?: (value: unknown) => TValue;
 };
 
-export function create(key: string, options?: ReferenceOptions) {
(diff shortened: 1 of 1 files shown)


jquense (approved):

jquense:
thanks!

dschaller:
Is this fix going to be available in 0.38.x?

PR #1295, opened 2021-03-10
feat: differentiate b/w positive and negative zero

ahmad2smile:
Details:

Changed following tests on `number` to consider `-0` and `0` and different number and to consider `-0 < 0`:

1. `number.min`
2. `number.max`
3. `number.lessThan`
4. `number.moreThan`

Essentially using ["Same Value"](https://tc39.es/ecma262/#sec-samevalue) algorithm used by `Object.is`, which specifically considers `-0` and `0`:

> This algorithm differs from the Strict Equality Comparison Algorithm in its treatment of signed zeroes and NaNs.

### DIFF
--- a/src/number.ts
+++ b/src/number.ts
@@ -43,14 +43,20 @@ export default class NumberSchema<
     return typeof value === 'number' && !isNaN(value);
   }
 
-  min(min: number | Reference<number>, message = locale.min) {
+  min(min: number | Reference<number>, message = locale.min) {    
     return this.test({
       message,
       name: 'min',
       exclusive: true,
       params: { min },
       test(value: Maybe<number>) {
-        return isAbsent(value) || value >= this.resolve(min);
+        const resolvedMin = this.resolve(min);
+
+        // NOTE: ES "Strict Equality Comparison" (i.e ===) does not differentiate b/w -0 and 0
+        // Ref: https://262.ecma-international.org/6.0/#sec-strict-equality-comparison
+        // "Object.is" implements "Same Value" algo which specifically considers -0 and 0
+        // Ref: https://tc39.es/ecma262/#sec-samevalue
+        return isAbsent(value) || Object.is(Math.min(value, resolvedMin), resolvedMin);
       },
     });
   }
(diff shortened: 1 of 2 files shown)


ahmad2smile:
Just FYI, this PR would make `0` and positive number indirectly, but does not change `number.positive` implementation, but a feature issue might rise users asking why not. and which was rejected in #322 

this is how it treats `0` as positive due to `Object.is`:

```js
Object.is(0, 0); // true
Object.is(0, +0); // true
Object.is(-0, +0); // false
Object.is(-0, 0); // false
```

src/number.ts
```diff
@@ -43,14 +43,20 @@ export default class NumberSchema<
     return typeof value === 'number' && !isNaN(value);
   }
 
-  min(min: number | Reference<number>, message = locale.min) {
+  min(min: number | Reference<number>, message = locale.min) {
```
thijskramer, line 46: there is some unnecessary whitespace here.

src/number.ts
```diff
@@ -43,14 +43,20 @@ export default class NumberSchema<
...
       exclusive: true,
       params: { min },
       test(value: Maybe<number>) {
-        return isAbsent(value) || value >= this.resolve(min);
+        const resolvedMin = this.resolve(min);
+
+        // NOTE: ES "Strict Equality Comparison" (i.e ===) does not differentiate b/w -0 and 0
+        // Ref: https://262.ecma-international.org/6.0/#sec-strict-equality-comparison
+        // "Object.is" implements "Same Value" algo which specifically considers -0 and 0
+        // Ref: https://tc39.es/ecma262/#sec-samevalue
```
thijskramer, line 58: Would it be possible to avoid abbreviations in your code? So please write `between` instead of `b/w`, and write `algorithm` instead of `algo`.

test/number.js
```diff
@@ -103,6 +103,15 @@ describe('Number types', function () {
       valid: [7, 35738787838, [null, schema.nullable()]],
       invalid: [2, null, [14, schema.min(10).min(15)]],
     });
+
+    it('should differentiate b/w -0 and 0', () => {
```
thijskramer, line 107: Could you write the correct word instead of `b/w`? I assume it means `between` but I first read it as `black/white` :-) 

Same goes for line 125, 150 and line 175.

Issue #1326, opened 2021-04-04
Returning a rejected Promise inside the test function results in an UnhandledPromiseRejection


danielrearden:
**Describe the bug**
Returning a rejected Promise inside the test function causes an unhandled Promise rejection

**To Reproduce**

See [this notebook](https://runkit.com/danielrearden/6069e094c9d81e001a0dec57) in RunKit for a reproduction of the issue.

**Expected behavior**
If the returned Promise is rejected, `validate` should also reject with the underlying error without raising an `unhandledRejection` exception. This is particularly problematic in Node 15 and on since unhandled rejections are [no longer just a warning](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V15.md#throw-on-unhandled-rejections---33021).

**Platform (please complete the following information):**
 - Node.js

**Additional context**
N/A

PR #1327, opened 2021-04-04
fix: prevent unhandled Promise rejection when returning rejected Promise inside test function

danielrearden:
Fixes https://github.com/jquense/yup/issues/1326


jquense (approved):

jquense:
Thanks!

Issue #1161, opened 2020-12-09
Anonymous arrow function warning with Next.js


k4rli:
**Describe the bug**

Warning message appears about anonymous arrow function. yup 0.32.5

```log
A codemod is available to fix the most common cases: https://nextjs.link/codemod-ndc
warn  - ./node_modules/yup/es/util/isSchema.js
Anonymous arrow functions cause Fast Refresh to not preserve local component state.
Please add a name to your function, for example:

Before
export default () => <div />;

After
const Named = () => <div />;
export default Named;

A codemod is available to fix the most common cases: https://nextjs.link/codemod-ndc

./node_modules/yup/es/util/isAbsent.js
Anonymous arrow functions cause Fast Refresh to not preserve local component state.
Please add a name to your function, for example:

Before
export default () => <div />;

After
const Named = () => <div />;
export default Named;

A codemod is available to fix the most common cases: https://nextjs.link/codemod-ndc
```

jquense:
This looks like a bug in Next, isSchema isn't a react component it shouldn't be checking it

tomdohnal:
I've come across this warning too and would like to revisit the issue.
1. I don't even if it's possible for libraries like Next.js to know is a function is or is not a React component.
2. I believe that changing 
```javascript
export default (value: any): value is undefined | null => value == null;
```
to
```javascript
const isAbsent = (value: any): value is undefined | null => value == null
export default isAbsent;
```
is not such a big deal. (Happy to send a PR)

Would you be willing to merge a PR with these changes? @jquense

PR #1329, opened 2021-04-08
fix(utils): use named functions for default exports

blurrah:
There's ESLint rules that warn on exporting an anonymous arrow function (See #1161) 

While I agree this does not break anything, it seems like a very small fix that does not impact yup and fixes a lot of console spam coming from NextJS (and maybe other projects that check for anonymous functions with fast refresh).


jquense (approved):

jquense:
thanks!

PR #1334, opened 2021-04-15
update lodash/lodash-es to fix CVEs flagged in 4.17.20

johnmccabe:
4.17.20 is flagged as being vulnerable to:

- CVE-2021-23337 Lodash versions prior to 4.17.21 are vulnerable to Command Injection via the template function.
- CVE-2020-28500 Lodash versions prior to 4.17.21 are vulnerable to Regular Expression Denial of Service (ReDoS) via the toNumber, trim and trimEnd functions.

See - https://www.openhub.net/p/lodash/security?filter%5Bmajor_version%5D=&filter%5Bperiod%5D=1&filter%5Bversion%5D=3409002148&filter%5Bseverity%5D=

Bumping to 4.17.21 for both lodash and lodash-es.

Previously the pinned versions for both drifted as the Lodash project had not been releasing lodash-es at the same time as lodash. They have resolved the release problems on their side and both are again released in sync.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,8 +110,8 @@
   "dependencies": {
     "@babel/runtime": "^7.10.5",
     "@types/lodash": "^4.14.165",
-    "lodash": "^4.17.20",
-    "lodash-es": "^4.17.15",
+    "lodash": "^4.17.21",
+    "lodash-es": "^4.17.21",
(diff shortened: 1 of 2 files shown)


jquense:
thanks!

PR #1305, opened 2021-03-22
Fix SchemaOf<>'s treatment of Date objects.

rustyconover:
Resolve #1243 and #1302.

Add a test for SchemaOf containing Date objects
and an array of date objects.

PR #1358, opened 2021-05-07
Allows custom types to be passed to avoid cast to ObjectSchema

cipacda:
Similar to  #1305, this allows the user to implement their own types for classes and not have fields forced to be of type ObjectSchema. This is very helpful when wanting to extend yup for ` js-joda LocalTime` for example.

This is backwards compatible as the new generic type `CustomTypes` is not mandatory and is defaulted to never.


jquense (approved):
o this is clever nice

PR #1427, opened 2021-07-23
feat: add types to setLocale

### LINKED ISSUE #1321: yup.setLocale() types
After upgrading the library and removing `@types/yup` to enable native TS support I've encountered a problem with `yup.setLocale()`. I am having issues with the following code:

```
yup.setLocale({
  string: {
    email: 'string.email',
    min: params => ({
      text: 'string.min',
      params,
    }),
    max: params => ({
      text: 'string.max',
      params,
    }),
  },
  number: {
    min: params => ({
      text: 'number.min',
      params,
    }),
    max: params => ({
      text: 'number.max',
      params,
    }),
  },
})
```

The approach is taken straight from the docs. So why is the type `setLocale(custom: any): void`? Won't `params.min` or `params.type` be present by default? Why those are not included. Yes, I can easily type it myself but it seems weird.

Are there plans for typing it and if not - why not?

Thanks!

maurodesouza:
The setLocale file was changed to .ts and the LocaleObject interface was created in locale.ts

Close #1321.

### DIFF
--- a/src/locale.ts
+++ b/src/locale.ts
@@ -52,6 +52,16 @@ export interface BooleanLocale {
   isValue?: Message;
 }
 
+export interface LocaleObject {
+  mixed?: MixedLocale;
+  string?: StringLocale;
+  number?: NumberLocale;
+  date?: DateLocale;
+  boolean?: BooleanLocale;
+  object?: ObjectLocale;
+  array?: ArrayLocale;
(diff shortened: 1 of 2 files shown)


jquense (approved):

jquense:
thanks!

PR #1435, opened 2021-08-05
Allow ref in oneOf/notOneOf

### LINKED ISSUE #1360: oneof with reference to array
I am having hard time to get this work. There is no much about ref in documentation and only example I found is password reference however I need to refer whole list which appears it is not working or am I doing something wrong?

here is my simple schema:
```
const validationSchema = object<SchemaFor<FormData>>({
  adminName: string().required(),
  categoryIds: array().min(1),
  mainCategoryId: string()
    .required()
    .oneOf([ref('categoryIds')])
});
```

despite I validate following data:
```
{
  "adminName": "Some Name",
  "categoryIds": [
    "cfd08de5-6ade-474d-be0c-2f806cda65db",
    "beee9dc9-c472-44e6-808a-1d8c896d8fe1"
  ],
  "mainCategoryId": "cfd08de5-6ade-474d-be0c-2f806cda65db"
}
```
I am still getting following error:
```
{
  "mainCategoryId": "mainCategoryId must be one of the following values: Ref(categoryIds)"
}
```
How ref should be correctly defined in this example?

akmjenkins:
It seemed like a neat idea. Closes #1360

### DIFF
--- a/src/util/ReferenceSet.ts
+++ b/src/util/ReferenceSet.ts
@@ -41,8 +41,11 @@ export default class ReferenceSet {
 
     let item,
       values = this.refs.values();
-    while (((item = values.next()), !item.done))
-      if (resolve(item.value) === value) return true;
+    while (((item = values.next()), !item.done)) {
+      const resolved = resolve(item.value);
+      if((Array.isArray(resolved) && resolved.some(r => r === value)) || resolved === value) return true;
+    }
+      
 
     return false;
   }
--- a/src/util/ReferenceSet.ts
+++ b/src/util/ReferenceSet.ts
@@ -1,5 +1,5 @@
-import Reference from '../Reference';
 import type { SchemaRefDescription } from '../schema';
+import Reference from '../Reference';
 
 export default class ReferenceSet {
   list: Set<unknown>;
--- a/test/mixed.js
+++ b/test/mixed.js
@@ -174,12 +174,13 @@ describe('Mixed Types ', () => {
   });
 
   it('should limit values with a ref', async () => {
-    let context = { someValues: [1,2,3] };
+    let someValues = [1,2,3];
+    let context = { someValues };
     let inst = mixed().oneOf(ref('$someValues'));
     await inst.validate(1,{context}).should.eventually.equal(1);
     await inst.validate(4,{context}).should.be.rejected().then(err => {
       err.type.should.equal('oneOf')
-      expect(err.params.resolved).to.deep.equal(context.someValues)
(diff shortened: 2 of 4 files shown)


test/mixed.js
```diff
@@ -173,6 +173,15 @@ describe('Mixed Types ', () => {
     );
   });
 
+  it('should limit values with a ref', async () => {
+    let context = { someValues: [1,2,3] };
+    let inst = mixed().oneOf(ref('$someValues'));
```
jquense, line 178: this is cool but i think it behaves differently than what someone would expect. I _ think_ what folks are asking for is that the array of values the ref resolves to be the list of valid values, as in 1, 2, or 3 would be valid values. That is a good bit harder to implement and i'm not sure that it'd be worth it

src/util/ReferenceSet.ts
```diff
@@ -41,8 +41,11 @@ export default class ReferenceSet {
 
     let item,
       values = this.refs.values();
-    while (((item = values.next()), !item.done))
-      if (resolve(item.value) === value) return true;
+    while (((item = values.next()), !item.done)) {
+      const resolved = resolve(item.value);
+      if((Array.isArray(resolved) && resolved.some(r => r === value)) || resolved === value) return true;
```
jquense, line 46: So this does work but it breaks the other case, where someone did "array().oneOf([ref('someArray')])` Admittedly that isn't very useful to do but it's confusing that array ref values are automatically flattened like this

akmjenkins: Interestingly, it does not break that case because of the `|| resolved == value`

Indeed, the following test passes:
```
    const r = [1,2,3];
    const m = array().oneOf([ref('$r')]).validateSync(r,{context:{r}});
    m.should.equal(r);
```

I could add it if you like.

But I also understand the hesitance to muddle this logic as well. It might seem more suitable to just instruct the user to add a custom test, which can easily be done in userland without further complicating the core.

jquense: I like this feature...I wonder if maybe we just tag the ref with `flatten = true` or something and then check for that in the ref set resolution logic?

akmjenkins: @jquense well, to be clear, due to #1437 getting merged, you can no longer do an equality comparison on the arrays - i.e. array().oneOf([ref('$someArray')]) - would no longer work due to the way concat works [here](https://github.com/jquense/yup/blob/master/src/util/ReferenceSet.ts#L30). 

Maybe I can fix that behaviour (allow the equality check) in this PR, and add the flatten flag as you suggest?

PR #1470, opened 2021-09-27
add an info about Transform feature.

MehdizadeMilad:
Transforming is somehow like Data Sanitization.


jquense:
hey thanks for the PR. Data sanitization is a use for transforms, but it's not really a good description of what 'transform' means, it may be a bit confusing to conflate these words in the docs

MehdizadeMilad:
True 👍
Thanks

Issue #1381, opened 2021-06-02
Unnecessary @babel/runtime dependency? 


jkimbo:
This is not really a bug report but more of a question: why does yup depend on `@babel/runtime` as a production dependency? I've downloaded the compiled library (https://registry.npmjs.org/yup/-/yup-0.32.9.tgz) and I can't find a reference to it.

Going down the babel compile chain it looks like `@babel/plugin-transform-runtime` is not enabled: 
https://github.com/jquense/yup/blob/877f7770ca54c0abea885230ffa307c1047226a3/rollup.config.js#L11
https://github.com/jquense/babel-preset-jason/blob/4fb48e83ed98359ec378ffb957e9395e7279b233/index.js#L86

so from my understanding of babel the runtime dependency is not required (caveat: I am by no means a babel expert so I might have missed something).

eternalmatt:
I am not a maintainer either, but I agree I think `@babel/runtime` should be moved to devDependencies unless it is used in the yup source code.

Definitely `@types/lodash` should be moved to devDependencies.

jkimbo:
@jquense I still don't think that `@babel/runtime` is needed for yup. I can't find a reference to it in the compiled source (https://registry.npmjs.org/yup/-/yup-0.32.11.tgz) and as far as I can see the flag to enable the `@babel/plugin-transform-runtime` plugin is not enabled in the .babelrc:
https://github.com/jquense/yup/blob/d072af33b84acef4f2bd355ff5637048edb95bae/.babelrc.js#L4-L15

PR #1483, opened 2021-10-11
Bug bash

jquense:


eternalmatt:
@jquense question about the `chore: bump deps`
1. Doesn't this undo the fix implemented in #486 and #488 ?
2. Please review https://github.com/jquense/yup/issues/1381 -- I think you have two dependencies that should be moved to devDependencies

PR #1512, opened 2021-11-06
docs(README): update block and allow terminology

matheusrufinos:
I would like to suggest a change in the README removing two words to lighter terms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,7 +604,7 @@ be used in the `message` argument.
 
 #### `mixed.oneOf(arrayOfValues: Array<any>, message?: string | function): Schema` Alias: `equals`
 
-Allowlist a set of values. Values added are automatically removed from any blocklist if they are in it.
+Only allow values from set of values. Values added are removed from any `notOneOf` values if present.
 The `${values}` interpolation can be used in the `message` argument. If a ref or refs are provided,
 the `${resolved}` interpolation can be used in the message argument to get the resolved values that were checked
 at validation time.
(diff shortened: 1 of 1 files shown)


jquense:
I'm happy to change arbitrary word choices in docs to be more clear and inclusive, but I do want to note that "whitelist" and "blacklist" are not "racist terms". they have no racial historic context or etymology. Unlike say master/slave which are clearly loaded terms (and have no place describe database replicas). I mention this because I don't imagine it's helpful to retroactively make words problematic or read weighty racial undertones into terms that don't have them. We have so much real insensitive words without creating new ones. 

I'm happy to change the terms here, tho I'd prefer "allow list" over "pass list"

README.md
```diff
@@ -604,7 +604,7 @@ be used in the `message` argument.
 
 #### `mixed.oneOf(arrayOfValues: Array<any>, message?: string | function): Schema` Alias: `equals`
 
-Whitelist a set of values. Values added are automatically removed from any blacklist if they are in it.
+Allowlist a set of values. Values added are automatically removed from any blocklist if they are in it.
```
jquense, line 607:
```suggestion
Allow a specific set of values. Values added are automatically removed from any blocklist if they are in it.
```

README.md
```diff
@@ -622,7 +622,7 @@ await schema.isValid(new Date()); // => false
 
 #### `mixed.notOneOf(arrayOfValues: Array<any>, message?: string | function)`
 
-Blacklist a set of values. Values added are automatically removed from any whitelist if they are in it.
+Blocklist a set of values. Values added are automatically removed from any allowlist if they are in it.
```
jquense, line 625:
```suggestion
Disallow values from a set of values. Values added are removed from `oneOf` values if present.
```

jquense:
thanks 👍

PR #1521, opened 2021-11-18
Fix documentation of mixed.strict

sheabunge:
[README.md](https://github.com/jquense/yup/#mixedstrictisstrict-boolean--false-schema) lists the default value of the `isStrict` argument of the `mixed.strict` function as `false`.

However, [schema.ts](https://github.com/jquense/yup/blob/master/src/schema.ts#L503) shows that the value used here is actually `true`. The documentation should be updated accordingly.


jquense (approved):

jquense:
thanks!

PR #1527, opened 2021-12-09
Add originalValue to TestContext type

### LINKED ISSUE #1335: Property `originalValue` does not exist on type `TestContext`
**Describe the bug**
Property `originalValue` does not exist on type `TestContext`.

**To Reproduce**

```javascript
.test(
  'is-valid-date',
  t('common:validation.date.format'),
  (value, context) => dayjs(context.originalValue).isValid())
```

On the above example, TypeScript complains that the `originalValue` prop doesn't exist on the `context`, although if you inspect the object, the prop is there.

**Expected behavior**
`originalValue` should be added to `TestContext`.

Moeface:
`originalValue` is listed as a valid property on `testContext` [per the docs](https://github.com/jquense/yup#mixedtestname-string-message-string--function-test-function-schema), and also exists on [the community types for yup](https://github.com/abnersajr/DefinitelyTyped/blob/a186d99d0c3a92424691a82130374a1b9145c7cd/types/yup/index.d.ts#L446), but it is missing in the official types.

Fixes #1335

### DIFF
--- a/src/util/createValidation.ts
+++ b/src/util/createValidation.ts
@@ -21,6 +21,7 @@ export type CreateErrorOptions = {
 export type TestContext<TContext = {}> = {
   path: string;
   options: ValidateOptions<TContext>;
+  originalValue: any;
   parent: any;
   schema: any; // TODO: Schema<any>;
   resolve: <T>(value: T | Reference<T>) => T;


jquense (approved):

jquense:
thanks!

PR #1590, opened 2022-02-15
notRequired description incorrect in README.md

BradWells:
The description for Schema.notRequired() reads that it is a shortcut to schema.notNull().defined() which seems the opposite of what is intended.

When I dug into the source I found this function.
`
notRequired(): any {
    return this.clone().withMutation((next) => {
      next.tests = next.tests.filter(
        (test) => test.OPTIONS!.name !== 'required',
      );
      return next.nullable().optional();
    });
  }
`


BradWells:
My edit might be wrong, still having trouble understanding if notRequired allows null and undefined, or just undefined.

If it allows just undefined then the alias is right but the description is still wrong.

PR #1594, opened 2022-02-23
🐛 Fix string.url() for localhost and ftp URL

### LINKED ISSUE #800: URL validation Failing for localhost
**Describe the bug**
Value http://localhost:8080/ return as a invalid url.

**To Reproduce**

```
const yup = require('yup');

const schema = yup.string().url();

const value = 'http://localhost:8080/'

console.log(`Testing '${value}'`, schema.isValidSync(value))
```

**Expected behavior**
This should return true.

**Platform:**
 - Chrome
 - Latest Version

### LINKED ISSUE #1271: string.url matcher is not passing on valid protocols/schemes
**Describe the bug**
`String.url` does not match other possible url protocols/schemas

**To Reproduce**

In the schema make a field a url matcher ` url: Yup.string().url()`
When validating a string that is not http/https/ftp it will fail

https://codesandbox.io/s/cool-cdn-pjk8k?file=/src/index.test.js

See [string.ts:11](https://github.com/jquense/yup/blob/master/src/string.ts#L11)

**Expected behavior**

 - Browser Chrome
 - Version 88

**Additional context**
The protocol (or scheme) can be a large variety of possible strings.
https://en.wikipedia.org/wiki/URL#Syntax

RuanAzevedo:
I had problems with `string.url()` validating `localhost` URLs, so that's the motivation of this PR.

I tried also to send a PR for the last stable release in npm `v0.32.11` (and that I'm using in a project), but this version's tests suite is broken. So, I'm sending the PR for the last version in the master branch.

Fix #1271 
Fix #800

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -153,10 +153,21 @@ export default class StringSchema<
       name: name || 'matches',
       message: message || locale.matches,
       params: { regex },
-      test: (value: Maybe<string>) =>
-        isAbsent(value) ||
-        (value === '' && excludeEmptyString) ||
-        value.search(regex) !== -1,
+      test: (value: Maybe<string>) => {
+        const regexTest =
+          isAbsent(value) ||
+          (value === '' && excludeEmptyString) ||
+          value.search(regex) !== -1;
+
+        if (regexTest) return regexTest;
+
+        try {
+          new URL(value);
+          return true;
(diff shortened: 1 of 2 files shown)


sysahmed:
HI This is spammer my email

PR #1636, opened 2022-04-15
fix: allow empty message for string.matches()

hduprat:
I wanted to return an empty error message on my fields. It worked with `string().required('')`, `string.min(10, '')` but not `string.matches(regex, '')` as it displayed the default error message.

This PR allows to have empty error messages for `string().matches(regex, '')` or `string().matches({message: ''})` (and subsequently, for `string().email('')`, `string().url('')` and `string().uuid('')`).

I also moved the default value of `message` and `name` in `matches()`, it doesn't cause a regression.


hduprat:
@jquense do you have time to review this pull request? If not, how can I be of help?

Issue #1633, opened 2022-04-13
Message type only accepts string | unknown, can Message type also accept ReactElement? 


feliciackh:
Hi, 

Currently the `required` function (as well as `email, URL, trim`, etc ) accepts a string or a function, can this be updated to accept `ReactElement` in the type definition? 

Currently we have:
```
const userSchema = yup.object({
  firstName: yup.string().min(2).required('First name is required'),
  lastName: yup.string().min(2).required('Last name is required'),
  email: yup.string().email('Valid email only'),
});
```

We have a function component that deals with translations for any text element and it returns a `ReactElement`, which is why we are trying to achieve the below:
```
const userSchema = yup.object({
  firstName: yup.string().min(2).required(<Format>First name is required</Format>),
  lastName: yup.string().min(2).required(<Format>Last name is required</Format>),
  email: yup.string().email(<Format>Valid email only</Format>),
});
```

Currently it works but it gives a type error on `Message` type. Is it possible to update Yup `Message` type definition to accept `ReactElement`, or is there a way to override the `Message` type definition?

Thanks!

jquense:
`unknown` should cover the the ReactElement case. I don't think i will add an explicit exception for the react element type, it would require taking a dependency on react types. Open to having the typing be more permissive there tho.

PR #1648, opened 2022-04-27
Allow consumers to configure additional types for Message

lexanth:
e.g. to allow adding ReactElement as an error message

Fixes https://github.com/jquense/yup/issues/1633

I tried to understand any performance impact of this
* From running typescript over just this repo many times over, there was no perceptible difference before/after this change
* Anecdotally, from applying this change as a patch to the types inside node_modules within our project, there was no obvious slow down.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -83,10 +83,33 @@ export interface MessageParams {
   spec: SchemaSpec<any> & Record<string, unknown>;
 }
 
+/**
+ * Placeholder to allow adding to the message union by type augmentation
+ *
+ * @example to add `ReactElement` to allowed message types
+ * ```ts
+ * declare module 'yup/lib/types' {
+ *   export interface CustomMessage extends BaseMessage {
+ *     type: ReactElement;
+ *   }
+ * }
+ * ```
+ * NB types cannot be removed from the message union with this, only added
+ */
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -83,10 +83,33 @@ export interface MessageParams {
...
+ *   export interface CustomMessage extends BaseMessage {
+ *     type: ReactElement;
+ *   }
+ * }
+ * ```
+ * NB types cannot be removed from the message union with this, only added
+ */
+export interface CustomMessage {}
+export interface BaseMessage {
+  type: unknown;
```
jquense, line 101: question: why is this approach better, vs extending the `Message` interface to be the type you want?

lexanth: As far as I know, you can't extend a union declared in another library (whereas you can augment interfaces, which this does do)

We (as a consumer) could have a wrapper around yup which defined `type OurMessage = Message | ReactElement`, but then we'd have to override all the yup types to accept it everywhere that yup's types use `Message` (or wrap all of yup etc).

lexanth:
@jquense any chance of getting this reviewed/merged?

danielantelo:
@jquense will you accept this change?

Issue #1610, opened 2022-03-13
Support Draft UUID Proposals


hackzilla:
**Describe the bug**
The change in #1112 to match the uuid regex exactly to the RFC 4122 broken my app.
I am using a draft version (6) of the uuid spec.

**To Reproduce**

**runnable** test case: https://codesandbox.io/s/peaceful-varahamihira-t5zd1h 

**Expected behavior**
Matching to allow anything that vaguely looks like a uuid.
Or potentially have a strict/relaxed uuid matching

**Platform (please complete the following information):**
 - Browser [all]
 - Version [0.32.9|master]

**Additional context**
There are a few new draft proposal which Yup is since #1112, treating as invalid.
https://github.com/uuid6/uuid6-ietf-draft

PR #1667, opened 2022-05-24
Offering relaxed uuid matching to support draft UUID proposals

wilsonvolker:
Rectifying the suggestions in #1610.

Added an addition regex to support relaxed uuid matching, which would allow anything that vaguely looks like an uuid. 
Also, updated the README.md.


wilsonvolker:
Also requested in #1705

PR #1675, opened 2022-05-27
Narrow type from array passed to mixed().oneOf()

### LINKED ISSUE #1230: TypeScript types: oneOf() should narrow/constrain the type
**Describe the bug**
`oneOf(..)` constraint should constrain the type to the narrowed values, but it doesn't.

**To Reproduce**
I couldn't make a repro with the codesandbox template because this is a TS-specific issue. However this snippet illustrates the issue:

```typescript
// Works, but overlayed broad type.
const unconstrained: Yup.SchemaOf<string> = Yup.string()
    .required()
    .oneOf(["foo", "bar"]);

// COMPILER ERROR OUTPUT:
// Type 'RequiredStringSchema<string | undefined, Record<string, any>>' is not assignable to type 'BaseSchema<Maybe<"foo">, Record<string, any>, "foo"> | BaseSchema<Maybe<"bar">, Record<string, any>, "bar">'.
//   Type 'RequiredStringSchema<string | undefined, Record<string, any>>' is not assignable to type 'BaseSchema<Maybe<"bar">, Record<string, any>, "bar">'.
//     Types of property '__inputType' are incompatible.
//       Type 'string | undefined' is not assignable to type 'Maybe<"bar">'.
//     Type 'string' is not assignable to type 'Maybe<"bar">'.
const broken: Yup.SchemaOf<"foo" | "bar"> = Yup.string()
    .required()
    .oneOf(["foo", "bar"]);
```

**Expected behavior**
Calling `oneOf()` should constrain the schema's type so that the `broken` value (see example above) compiles. **I'm also open to other ways of accomplishing this if I'm just "doing it wrong"**.

The source (for master at the time of this issue) is part of `BaseSchema`:
https://github.com/jquense/yup/blob/67c96ae7f8fcb59758732b2956ff0b7390f08213/src/schema.ts#L668-L697

...However, I don't think there's a way to constrain it on `BaseSchema` since it uses the polymorphic `this` return type, is there? I think you could override the method in `StringSchema`, etc to specify a stricter return type... kind of awkward, but it would be an improvement. Let me know your thoughts about a solution - **I can make a PR implementing it** if you'd like. I've made a few for you in the past so you know I'm good for it.

Current version of Yup (0.32.8). Problem is also present in master as of now (https://github.com/jquense/yup/tree/67c96ae7f8fcb59758732b2956ff0b7390f08213).

**Additional context**
Issue #1165 looks to be related, but there isn't any mention of `as const` in the readme anymore so I'm guessing it's out of date/not relevant anymore.

karlhorky:
Closes #1230

### DIFF
--- a/src/mixed.ts
+++ b/src/mixed.ts
@@ -49,6 +49,11 @@ export default interface MixedSchema<
     def: Thunk<D>,
   ): MixedSchema<TType, TContext, D, ToggleDefault<TFlags, D>>;
 
+  oneOf<U extends TType>(
+    arrayOfValues: ReadonlyArray<U | Reference>,
+    message?: MixedLocale['oneOf'],
+  ): MixedSchema<U, TContext, TDefault, TFlags>;
+
   concat<IT, IC, ID, IF extends Flags>(
     schema: MixedSchema<IT, IC, ID, IF>,
   ): MixedSchema<Concat<TType, IT>, TContext & IC, ID, TFlags | IF>;


karlhorky:
@jquense friendly bump, what do you think of the approach of this PR?

karlhorky:
@jquense closed #1230 just a couple of days ago: https://github.com/jquense/yup/issues/1230#issuecomment-1223143895

So it's possible this has been implemented in `1.0.0-beta.5` or `1.0.0-beta.7`

jquense:
Mixed still doesn't narrow, but string does. I'm not convinced that mixed should narrow, it may make the type too restrictive. Still need to think about it a bit

karlhorky:
> but string does

Is this new in `1.0.0-beta.5` or `1.0.0-beta.7`? What does this look like in practice?

---

> I'm not convinced that mixed should narrow

To me, it seems like usage of `mixed().oneOf()` should narrow in all cases. Especially if using 

For example, in what case would a user want something different than `'abc' | 'def'` as the type of the following?

```ts
const mixed = yup.mixed().oneOf(['abc', 'def'] as const);
```

karlhorky:
Regardless of whether it should narrow, `mixed().oneOf()` at least not return `any` as it currently does, see [this sandbox](https://codesandbox.io/s/blissful-field-xw9kjn?file=/src/index.ts) for a demonstration of very broken behavior:

```ts
import { mixed } from "yup";

type IsAny<T> = unknown extends T ? (T extends {} ? T : never) : never;

type NotAny<T> = T extends IsAny<T> ? never : T;
function notAny<T>(x: NotAny<T>) {}

// Without using .oneOf() retains the type of MixedSchema<null, AnyObject, undefined, ''>
const schemaWorking = mixed<null>().defined();
notAny(schemaWorking); // ✅ This type is not `any`

// Using .oneOf() sets the type to `any`
const schemaBroken = mixed<null>().oneOf([null]).defined();
notAny(schemaBroken); // 💥 Type is `any`
```

jquense:
>  Is this new in 1.0.0-beta.5 or 1.0.0-beta.7? What does this look like in practice?

image

not new, its been in in most of the v1 releases

> Regardless of whether it should narrow, mixed().oneOf() at least not return any as it currently does,

that is definately true, will fix

slashwhatever:
Update on this?

karlhorky:
@jquense do you think that you would reconsider this PR?

Would be great to be able to represent this properly:

```ts
interface FormidableFile {
  hashAlgorithm: false | "sha1" | "md5" | "sha256";
}

const formidableFileSchema: ObjectSchema<FormidableFile> = object({
  hashAlgorithm: mixed()
    .oneOf([false, 'sha1', 'md5', 'sha256'] as const)
    .required(),
})
```

Currently this returns the following (confusing) error:

```
Type 'ObjectSchema<{ hashAlgorithm: AnyPresentValue; }, AnyObject, { hashAlgorithm: undefined; }, "">' is not assignable to type 'ObjectSchema<FormidableFile, AnyObject, any, "">'.
  The types of 'default(...).fields.hashAlgorithm' are incompatible between these types.
    Type 'Reference<unknown> | ISchema<AnyPresentValue, AnyObject, any, any>' is not assignable to type 'Reference<unknown> | ISchema<false | "sha1" | "md5" | "sha256", AnyObject, any, any>'.
      Type 'ISchema<AnyPresentValue, AnyObject, any, any>' is not assignable to type 'Reference<unknown> | ISchema<false | "sha1" | "md5" | "sha256", AnyObject, any, any>'.
        Type 'ISchema<AnyPresentValue, AnyObject, any, any>' is not assignable to type 'ISchema<false | "sha1" | "md5" | "sha256", AnyObject, any, any>'.
          Type 'AnyPresentValue' is not assignable to type 'false | "sha1" | "md5" | "sha256"'.
```

PR #1698, opened 2022-06-26
Standard RFC compliance email validation

yasirshahzad:
Hello, I hope you are doing well. 

The current regex for validating emails is not validating correctly for the following cases. 

- “example”@email.com
- “obviously”not”correct@email.com
- example…example@email.com
- CAT…123@email.com
- おえあいう@example.com
- example@111.222.333.44444

**Problem**
Current regex doesn't validate these cases. 

**Solution:** 
I have added a new regex for email validation that is supported by RFC. The supportive links are given below. 

**References:** 
- https://datatracker.ietf.org/doc/html/rfc3696
- https://datatracker.ietf.org/doc/html/rfc5321
- https://datatracker.ietf.org/doc/html/rfc5322

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -18,7 +18,7 @@ import Schema from './schema';
 
 let rEmail =
   // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


hewseph:
Anyone gonna merge this? I'm tired of writing email regex for `yup`

Svish:
From the documentation of [string().email() validation](https://github.com/jquense/yup#stringemailmessage-string--function-schema):

> Validating email addresses is nearly impossible with just code. Different clients and servers accept different things and many diverge from the various specs defining "valid" emails. The ONLY real way to validate an email address is to send a verification email to it and check that the user got it. With that in mind, yup picks a relatively simple regex that does not cover all cases, but aligns with browser input validation behavior since HTML forms are a common use case for yup.

Right below that, it also notes how you can override this, if you for some reason really need that:
```ts
yup.addMethod(yup.string, 'email', function validateEmail(message) {
  return this.matches(myEmailRegex, {
    message,
    name: 'email',
    excludeEmptyString: true,
  });
});
```

jquense:
Yes this PR predates the lastest change but it does mean it's obsolete now. Thanks for taking the time to contribute, sorry for the extremely slow response on my end

PR #1711, opened 2022-07-06
Fix README.md

nandorojo:
The current docs recommend destructuring as an array here. However, this throws an error, since it's returning the value when you pass `'isBig'` as a string directly.


jacobdo2:
For `v0.32.11` there is a mismatch in `.when` between type definition and the behaviour of the method.

See codesandbox here: https://codesandbox.io/s/romantic-matsumoto-gsj34u?file=/src/index.ts

Seems that types that go with this version are also intended for v1.

Issue #768, opened 2020-02-01
`OneOf` and `notOneOf` do not handle `null` and `undefined` in the error message.


DaddyWarbucks:
**Describe the bug**
When using the `oneOf` or `notOneOf` with `null` and `undefined` the values are not interpolated into the error message properly.

```js
const oneOfSchema = yup.object({
    name: yup
      .string()
      .nullable()
      .oneOf([null, undefined])
  });

// "name must be one of the following values: , "

// ^^ Notice the `null` and `undefined` are not "stringified" into the error message.
```

**To Reproduce**

https://codesandbox.io/s/dreamy-elgamal-7d2fv

**Expected behavior**
When using `null` or `undefined` in the `oneOf` or `notOneOf` methods the error message will interpolate those values correctly.

**Platform (please complete the following information):**
 - Chrome 79 on macOS (Mojave)
 - Version: 0.26.10

**Additional context**
This may seem silly that someone would mark a schema as `nullable` but then also set the schema as `oneOf([null])`, but I have a use case where when using `when` I believe this makes sense.

```js
stripe_id: yup
      .string()
      .label('Stripe ID')
      .required()
      .nullable()
      .when('type', (type, schema) => {
        // If type is stripe, stripe_id cannot be null and must be provided
        // Otherwise the value must be null
        return type === 'stripe'
          ? schema.nullable(false)
          : schema.oneOf([null]);
      })
```

Issue #104, opened 2017-07-14
mixed.oneOf doesn't respect nullable


fnky:
The [`mixed.oneOf`](https://github.com/jquense/yup#mixedoneofarrayofvalues-arrayany-string-message-schema-alias-equals) doesn't accept `null` when `nullable` is added.

##### Example

```js
const fruitsOrNone = yup.string().oneOf(['apple', 'banana']).nullable(true)

fruitsOrNone.validate(null) // => throws ValidationError: 'this must be one the following values: apple, banana'
fruitsOrNone.validate('apple') // => apple
```

##### Expected

It's expected to accept the value `null` when `nullable` is set on the schema.

jquense:
yes this confusing but expected. you have a contradiction between your schema on one hand you are saying its can _only_ be 'apple' or 'banana' and on the other say it can also be `null`, the best way to handle this is to add `null` to your oneOf config

aphillipo:
I strongly think that nullable() should add the null to the oneOf constraint automatically.  Sometimes you are reusing the options passed to oneOf for your dropdowns (for example) but you don't want a null option field added.

I'm currently doing ```.oneOf([null].concat(RELATIONSHIPS), message)``` or whatever fields you have in there and it looks mega weird!

nfantone:
> yes this confusing but expected. you have a contradiction between your schema on one hand you are saying its can only be 'apple' or 'banana' and on the other say it can also be null

@jquense I could make the same argument for every schema definition that includes `.nullable()`. For instance, by using `yup.number().nullable()` you are saying that "this is confusing" because "you have a contradiction" by saying that your value "can _only_ be a number" and "on the other side, it says in can also be `null`". The point is: _yes_, that is exactly what I'm saying it should happen. 

This boils down to a matter of semantics and expressing intent. I honestly wouldn't find it confusing at all to read `yup.string().oneOf(['apple', 'banana'
]).nullable()` (so, `"apple"`, `"banana"` and `null` are all fine) - in the same way, I wouldn't think `yup.number().nullable()` is contradictory or unexpected.

jquense:
[...]

PR #1757, opened 2022-08-22
feat: respect nullable() with oneOf

jquense:
closes: #768 #104

BREAKING CHANGE: previously `oneOf` required adding `null` explicitly to allowed values when using oneOf. Folks have found this confusing and unintuitive so I am deferring and adjusting the behavior

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -848,16 +848,16 @@ export default abstract class Schema<
     next.internalTests.whiteList = createValidation({
       message,
       name: 'oneOf',
+      skipAbsent: true,
       test(value) {
-        if (value === undefined) return true;
-        let valids = this.schema._whitelist;
+        let valids = (this.schema as Schema)._whitelist;
         let resolved = valids.resolveAll(this.resolve);
 
         return resolved.includes(value)
           ? true
           : this.createError({
               params: {
-                values: valids.toArray().join(', '),
+                values: Array.from(valids).join(', '),
                 resolved,
               },
             });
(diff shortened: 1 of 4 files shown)


karlhorky:
Thanks for this @jquense ! 

How do you expect users to represent "only `null`" using `mixed()` now? My previous solution here does not seem to work anymore:

Screenshot 2022-11-12 at 10 46 35


```ts
  }).shape({
    files: mixed()
      .label('Files')
      .oneOf([null], ({ label }) => `${label} must be null`)
      .nullable()
      .defined(),
  });
```

Error message:

```
Type 'null' is not assignable to type 'Reference<unknown> | AnyPresentValue | undefined'.
```

karlhorky:
Oh, maybe just moving `.nullable()` up one line? This seems to not fail:

```diff
  }).shape({
    files: mixed()
      .label('Files')
+      .nullable()
      .oneOf([null], ({ label }) => `${label} must be null`)
-      .nullable()
      .defined(),
  });
```

jquense:
simplest way to do this is `const nullSchema = mixed((value) === null)`

karlhorky:
Hm, do you mean the following with a type predicate? This seems to throw a type error.

```ts
mixed((value): value is null => value === null) // 💥 Type 'null' is not assignable to type 'AnyPresentValue'.
  .nullable();
```

Full TypeScript error:

```
Argument of type '(value: any) => value is null' is not assignable to parameter of type 'MixedOptions<AnyPresentValue> | TypeGuard<AnyPresentValue> | undefined'.
  Type '(value: any) => value is null' is not assignable to type 'TypeGuard<AnyPresentValue>'.
    Type predicate 'value is null' is not assignable to 'value is AnyPresentValue'.
      Type 'null' is not assignable to type 'AnyPresentValue'.
```

PR #1782, opened 2022-09-22
toString Bug Fixed

andreydiskin:
Fixed a little bug - 
string validation allow int typed input to be inserted, causing wrong typing in outcome data,
const strValue = value != null && value.toString ? value.toString() : value; (string.ts file, line 82)


jquense:
Do tests pass with this? This would not parse numbers to strings, which is the intended behavior

PR #1810, opened 2022-10-16
Empty string bug added by SpaceBar fixed

Muhammad-zain-rehman:
During Validation of any form Press the Space bar and submit a form, Form will be Submit, Empty String Should Validate False,
I always write a custom method to validate String with the Required method whenever I use Yup for validation. Today I just thought about fork this Repo and want to contribute, if there is any wrong Please let me know. I'm a beginner.

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -94,6 +94,29 @@ export default class StringSchema<
       }),
     );
   }
+  
+  /*
+  CommentId: Zain:#1 (open)
+  Bug Detail: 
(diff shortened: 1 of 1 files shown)


jquense:
hey there, thanks for the PR. I think this isn't something we would add to the base string type. it's a bit specific to use-cases and it's also fairly straightforward to implement with the existing base methods and validations. Overall its valuable just not worth including for everyone who uses yup out of the box in terms of added bytes

PR #1811, opened 2022-10-17
Fix addMethod base type.

rosskevin:
I also note that there are no ci/automated tests for the types file.  I'm not sure what the plan is here, but I do see that the conversion of the `addMethod` test from `yup.js` could be challenging in that regard because it is dynamic.  ~~Perhaps it is best to get a typecheck done on this `test/types/types.ts` file if there are no other solutions?  Something like https://github.com/i18next/i18next/blob/master/package.json#L109 or better?~~ I see type checking is done via jest.

Regardless, this PR gets `addMethod` into a working state on the 1.0.0-beta.7.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,12 +27,12 @@ import Schema, {
 } from './schema';
 import type { InferType, Message } from './types';
 
-function addMethod<T extends Schema>(
+function addMethod<T extends AnySchema>(
   schemaType: (...arg: any[]) => T,
   name: string,
   fn: (this: T, ...args: any[]) => T,
 ): void;
-function addMethod<T extends new (...args: any) => Schema>(
+function addMethod<T extends new (...args: any) => AnySchema>(
   schemaType: T,
   name: string,
   fn: (this: InstanceType<T>, ...args: any[]) => InstanceType<T>,
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,12 +27,12 @@ import Schema, {
 } from './schema';
 import type { InferType, Message } from './types';
 
-function addMethod<T extends AnySchema>(
+function addMethod<T extends Schema>(
   schemaType: (...arg: any[]) => T,
   name: string,
   fn: (this: T, ...args: any[]) => T,
 ): void;
-function addMethod<T extends new (...args: any) => AnySchema>(
+function addMethod<T extends new (...args: any) => Schema>(
   schemaType: T,
   name: string,
   fn: (this: InstanceType<T>, ...args: any[]) => InstanceType<T>,
--- a/test/types/types.ts
+++ b/test/types/types.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable no-unused-labels */
 import {
+  addMethod,
   array,
   number,
   string,
@@ -1035,4 +1036,14 @@ reach: {
 
   // $ExpectType Reference<"foo"> | ISchema<"foo", AnyObject, any, any>
   const _3 = reach(obj, 'ref');
+}
+
+AddMethod: {
+  // NOTE: if this changes, update the doc https://github.com/jquense/yup#extending-built-in-schema-with-new-methods
+  addMethod(string, 'append', function append(appendStr: string) {
+    return this.transform((value) => `${value}${appendStr}`);
+  });
+  addMethod(string, 'mixed', function append(appendStr: string) {
+    return this.transform((value) => `${value}${appendStr}`);
+  });
 }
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -27,12 +27,12 @@ import Schema, {
 } from './schema';
 import type { InferType, Message } from './types';
 
-function addMethod<T extends AnySchema>(
+function addMethod<T extends Schema>(
```
rosskevin, line 30: This could be invalid, and the necessary change could be that `AnySchema` is no longer correct.  I expect `AnySchema` to allow _any schema_ to be passed, but it appears that is not the case. I'm having a hard time making sense of the class hierarchy vs interface hierarchy, so this PR (the addMethod signature change) may be invalid, though I would like to see the change to test/types.ts added.

rosskevin: @jquense if you could give me some background or direction, I can alter the PR.  At this point in my code updating to 1.0.0-beta, I think that `AnySchema` is no longer fulfilling its purpose.  Please let me know...

rosskevin: `reach()` results in `ISchema<any, AnyObject, any, any>` so I'm even more confused about the base interface.

jquense: AnySchema is `Schema<any, any, any>`, e.g. with the generics filled in. Its generally what you want when assigning to a thing that inherits `Schema`, b/c otherwise you get the default schema generic values which aren't always assignable to any other schema. At a high level they are the same type but `AnySchema` is more permissive about the inner generics.

That said, there are other "schema like" classes that do not extend Schema, in particular `Lazy`, so `ISchema` is the common lowest interface to describe anything with with a type, and cast and validate methods. Reach types are not great, but they return `ISchema` b/c reach can return a Lazy instance as well as a Schema

Overall, I think AnySchema is still correct here.

rosskevin: Agreed, I think in purpose `AnySchema` is the correct idea, but it fails with:

```
No overload matches this call.
  Overload 1 of 2, '(schemaType: (...arg: any[]) => AnySchema<any, AnyObject, any, Flags>, name: string, fn: (this: AnySchema<any, AnyObject, any, Flags>, ...args: any[]) => AnySchema<...>): void', gave the following error.
    Argument of type 'typeof create' is not assignable to parameter of type '(...arg: any[]) => AnySchema<any, AnyObject, any, Flags>'.
      Call signature return types 'StringSchema<string | undefined, AnyObject, undefined, "">' and 'AnySchema<any, AnyObject, any, Flags>' are incompatible.
        The types of '__outputType' are incompatible between these types.
          Type 'string | undefined' is not assignable to type 'never'.
            Type 'undefined' is not assignable to type 'never'.
  Overload 2 of 2, '(schemaType: new (...args: any) => AnySchema<any, AnyObject, any, Flags>, name: string, fn: (this: AnySchema<any, AnyObject, any, Flags>, ...args: any[]) => AnySchema<...>): void', gave the following error.
    Argument of type 'typeof create' is not assignable to parameter of type 'new (...args: any) => AnySchema<any, AnyObject, any, Flags>'.
      Type 'typeof create' provides no match for the signature 'new (...args: any): AnySchema<any, AnyObject, any, Flags>'.
```

I have pushed the test which uses the `addMethod` sample from the README to demonstrate the error. I see other errors in my codebase that are similar, making me thing a modification to `AnySchema` may be necessary, but for the life of me I haven't been able to get one to work out.

rosskevin: Actually, for some reason I've got clarity on this, and I'm going to revert to my original thought.  `AnySchema` doesn't make sense here because we are using `<T extends`, and `StringSchema` extends `Schema`, not `AnySchema`. So it seems the use of `AnySchema` with `<T extends` may be the incorrect thing to do. I'm going to try this out and see.

rosskevin: Some observations:
- `Schema` is already parameterized as `any`, with the exception of `Flags` being defaulted differently.
- `abstract class Schema implements ISchema` but not `Schema`, while none of the other class implementations `implement` their same-named interfaces e.g. `class StringSchema extends Schema`.  So why bother with declaring same-named interfaces?  This has led me to be more confused than I think is necessary.

I'm game to help and PR some clarifications  - I'm just trying to make sure I understand some of the intent and design decisions.  I'm happy to take direction and pitch in.

rosskevin:
This is not so much a criticism, because I'm very thankful for your work, but a summary of my thinking after spending a couple of hours looking at the code:

I've been experimenting with a few things, and I'm having a hard time following same-named classes and interfaces, one not implementing the other, and general hierarchies that don't extend each other.  It may just be the style of writing, but my brain comes to a screeching halt trying to mentally index what is an interface and what is a class.  I'm just going to bail out.  

I'm game to help out in the future, but the current state of the code is not something I can track with any clarity.

I suggest (but by no means expect you to follow these suggestions):
- solidifying an interface hierarchy
- classes implement interfaces
- rename `ISchema` -> `SchemaLike`
- add the general intent/purpose to the interface header docs (this would definitely help outsiders stick to intent)
- perhaps one way to avoid confusion would be to differentiate class vs interface name, and only used named exports

Again, I have no expectation, this is just feedback from someone on the outside looking to help.  Thanks again for your work.

PR #1812, opened 2022-10-17
Export ValidateOptions for external reuse

rosskevin:
We have been using this type from the general release, and I see no reason to hide it, given that `InternalOptions` serves this purpose.


rosskevin:
I suggest this is a wormhole, and unless necessary, use `export * from`.  
index_ts_—_js

Is there any reason to hide types?

jquense (approved):

jquense:
thanks!

Issue #1814, opened 2022-10-21
No info about addMethod for this.test()


Mikilll94:
It's not described how to use `addMethod` for  `this.test()` in the README.md.
https://github.com/jquense/yup#addmethodschematype-schema-name-string-method--schema-void

Currently if you want to add `this.test()` method to the schema you need to google for that because there is completely no info about this in the docs.
https://medium.com/@iamarkadyt/how-does-yup-addmethod-work-creating-custom-validation-functions-with-yup-8fddb71a5470
https://snyk.io/advisor/npm-package/yup/functions/yup.addMethod
https://stackoverflow.com/a/66758718/6182181

brianwhaley:
Starting the work on this...

brianwhaley:
I provided details and an explicit example where addmethod can be a test.  

Just submitted the PR, you can see it there.  https://github.com/jquense/yup/pull/1817

Mikilll94:
@jquense 

This syntax is completely undocumented

```ts
Yup.addMethod(Yup.type, ‘methodName’, function(anyArgsYouNeed) {
    const { message } = anyArgsYouNeed;
    return this.test(‘test-name’, message, function(value) {
        const { path, createError } = this;
        const { some, more, args } = anyArgsYouNeed;
        return createError({
            path,
            message: "This field is invalid."
        });
    });
});
```

jquense:
Sorry, I'm still not sure what you mean. The docs do not show every possible use case but all of that is documented. [here](https://github.com/jquense/yup#addmethodschematype-schema-name-string-method--schema-void) and [here](https://github.com/jquense/yup#schematestoptions-object-schema)

Mikilll94:
@jquense
1. `yupe.addMethod` arguments are not documented.
2. Adding new test method is not documented.
3. `this.test` method and its arguments are not documented.
4. Elements inside `this` object, i.e. `path`, `createError` are not documented.

jquense:
@Mikilll94 all of those things are documented, I linked the documentation above for most of them. I think you may be unfamiliar with how instance methods work on JS classes? `this` is not special, it is the the schema, schema are all class instances. `string().test()` is the same a `this.test` here.

[...]

PR #1817, opened 2022-10-28
added info about addMethod for this.test()

brianwhaley:
Updated README.md to add info about addMethod for this.test() re: https://github.com/jquense/yup/issues/1814

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,19 @@ import { addMethod, Schema } from 'yup';
 addMethod(Schema, 'myMethod', ...)
 ```
 
+
+You can add your test methods to the schema using `addMethod` by returning `this.test()`.  This will allow you to reuse them as well. 
+
+```ts
+Yup.addMethod(Yup.mixed, 'myLength', function(myLength, msg) {
+   return this.test({
+      name: 'myLength',
+      message: msg,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -566,6 +566,19 @@ import { addMethod, Schema } from 'yup';
 addMethod(Schema, 'myMethod', ...)
 ```
 
+
+You can add your test methods to the schema using `addMethod` by returning `this.test()`.  This will allow you to reuse them as well.
```
jquense, line 570: Rather than add add a new section here, let's add the code example to the first code block and then demonstrate how to use the new added methods. For extra clarity i think it's worth adding a few words about `this` being the the schema instance, and so any of the methods documented in the docs will be on them.

brianwhaley: Sure.  I will make the change and resubmit.

Svish:
Would be much better if this example was in Typescript, and included what one would need to add to the `.d.ts` to make things work properly.

zammit123:
Yes a Typescript example would be very helpful. Please add one.

tranlehaiquan:
It's good for document, does it need something else to merge? @jquense?

PR #1818, opened 2022-10-29
Empty transform

HamzaM3:
Added a simple empty transform. It just take an array of values or a single value and turns it into undefined (within the transforms reduce).

It would be better if the values given would be added to a new property and if there's a check before the transforms reduce in order to launch getDefault directly.

What do you think ?


HamzaM3:
Okay yes I understand, thank you for your explanation. I'll try to follow this logic from now.

Thank you very much

PR #1824, opened 2022-11-07
fix: string().notRequired()

lwassermann:
In a project, we need to undo some `.required()` fields in `.when()`-conditions, but for `string()` that does not work 🙈.

This should fix it. 🤞 

Thank you for your work maintaining this validation library (coming here from the Formik-recommendation). 💚


jquense (approved):

jquense:
Thanks!

PR #1828, opened 2022-11-15
Fix example

karlhorky:
It appears that `mixed().null` does not exist, TypeScript error:

```
Property 'null' does not exist on type 'MixedSchema<AnyPresentValue | undefined, AnyObject, undefined, "">'.
```


jquense (approved):

jquense:
thanks

karlhorky:
Glad to help, thanks for the merge!

Issue #1400, opened 2021-06-21
yup.string().trim() throw error when value is not string


cstria0106:
Thank you for awesome library. I'm sorry for my poor english.

**Describe the bug**
When value is not string, yup.string().trim() throws error.
```sh
(node:8353) UnhandledPromiseRejectionWarning: TypeError: val.trim is not a function
```

**To Reproduce**

```ts
yup.object({
    name: yup.string().trim(),
}).validateSync({
    name: {
        first: 'firstname',
        last: 'lastname',
    },
});
```

**Expected behavior**
I expected validationSync to be failed because value.name is not string.

**Platform (please complete the following information):**
 - Node.js
 - v14.16.0
 - yup
 - v0.32.9

**Additional context**
I found that the error not occurs in version 0.27.0 (https://codesandbox.io/s/yup-test-case-gg1g1) (in converts object to '[object Object]')

thudaohd:
I have same issue

Arieski:
Same issue here

sysahmed:
Spam

AnechaS:
You can specify the flag option `strict`: true

```js
yup
  .object({ name: yup.string().trim().nullable() })
  .validateSync(
    {
      name: {
        first: 'firstname',
        last: 'lastname',
      },
    },
    {
      strict: true,
    }
  );
```

PR #1831, opened 2022-11-17
trim only when type is string

sagarpanchal:
Fixes https://github.com/jquense/yup/issues/1400


sagarpanchal:
@jquense Please, check this.
I don't know what was the reason for checking null instead of checking type. Perhaps, performance?
But, I have placed type checks before running trim(). So, it doesn't break.

PR #1843, opened 2022-11-25
Number shouldnt accept arrays

HamzaM3:
This is crazy, but parseFloat casts arrays as well !!


HamzaM3:
Could you please tell me what is the exact behavior expected here. Do we want to cast strings and that's it ? Because I understand that parseFloat does a toString before casting. So we are casting anything that has a toString that starts with a number ! Does the user expect this ?

PR #1860, opened 2022-12-16
Fix error in documentation

Strato:
`NotRequired` applies both `nullable` and `optional`. This is confirmed by the code itself :

```ts
notRequired(): any {
  return this.clone().withMutation((next) => next.nullable().optional());
}
```

It's also **not** an alias of `optional` since it applies `nullable` as well.


jquense (approved):
thanks!

PR #1866, opened 2022-12-26
`value` param as optional in `.cast`

bluelovers:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40683


jquense:
Hi, thanks for the PR. I'm going to leave the type as it is, calling cast without any arguments isn't actually correct, and while its functionally the same as `schema.cast(undefined)` it's less clear. If you want to generate default values that is what `getDefault()` is for

Issue #1886, opened 2023-01-21
Improvement: Add result of getDefault() to describe


DaddyWarbucks:
Is there any reason that the `describe()` method does not return the default value? It seems like it would be trivial to add a `getDefault()` it to this code: https://github.com/jquense/yup/blob/bc74c340721da2ea6e65cb27b967c2970af44d35/src/schema.ts#L912

I am happy to work on this, but I am just wondering if its not already done for a reason.

DaddyWarbucks:
Closing because the corresponding PR was merged. Thanks!

PR #1887, opened 2023-01-21
Add getDefault results to the describe method

DaddyWarbucks:
This PR addresses https://github.com/jquense/yup/issues/1886.

The `describe()` method now returns the result of `schema.getDefault()`. I am currently updating a library that converts Yup schemas to JSONSchemas. The new `optional` and `nullable` flags have been helpful. The addition of the `default` would be super helpful as well.

Let me know if you have any questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,7 +681,7 @@ let schema = object({
 schema.describe({ value: { isBig: true } });
 ```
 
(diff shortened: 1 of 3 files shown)


jquense (approved):
Nice, thanks

Issue #1906, opened 2023-02-08
How do I migrate to v1


jquense:
Please read through the releases: https://github.com/jquense/yup/releases which document the breaking changes in more detail but the **tl;dr;** here is the following:

## Big additions

The `tuple` type, object `omit`, `pick`, `partial` and `deepPartial` better TS support, most consistent behavior and 

### Flat bundle

There is no distributed `lib` folder anymore everything is bundled into one file with rollup

### Stricter `when` API

https://github.com/jquense/yup/pull/1542

### Nullability and optionality are stricter and enforced during `cast`

This is the largest, and likely most disruptive change. Prior yup allowed for patterns like:

```ts
const nullableRequiredString = string().nullable().required()

nullableRequiredString.cast(null) // -> null

nullableRequiredString.validate(null) // ValidationError("this is required and cannot be null")
```

This may seem unintuitive behavior (and it is) but allowed for a common client side validation case, where we want to use a single schema to parse server data, as well as validate user input. In other words, a server might return invalid "default" values that should still fail when trying to submit.

Now, `nullable()`, `defined` and `required` are all mutually dependent methods. Meaning `string().nullable().defined().required()` produces a schema where the value must be a string, and not null or undefined. The effect of this is that **the type of a cast() is now accurate and the same as the type returned from validate.**

**There is a migration path** for folks who use this pattern: `name.cast(null, { assert: 'ignore-optionality'}) // => null` read more [here](https://github.com/jquense/yup/releases/tag/v1.0.0-beta.5)

### TS generics are faster, more accurate and very different from v032.

`SchemaOf<Person>` has been removed, and you can now provide the type you want the schema to match: `ObjectSchema<Person>`. read more: [here](https://github.com/jquense/yup/releases/tag/v1.0.0-alpha.3)

frck006:
Hi,

I found nothing about 'BaseSchema' from 'yup', it disapears in v1 ? We can replace it by something else ?

Regards,

jquense:
Yes it's just Schema now

anonmily:
Hi, 

We've switched to yup v1, and it seems to have broken the type imports since lib doesn't exist any longer. For v1 how should we go about accessing the types like MessageParams and ConditionOptions? They seem to be in the yup index.d.ts file, but don't seem to be exported in the yup module, so I'm unsure about how to access the types with v1.

jquense:
If you need more types exported open an issue (or better a PR)

gajus:
[...]

Issue #1579, opened 2022-02-07
1.0 beta includes breaking changes for .when beyond just function version


evanseeds:
**Describe the bug**
1.0.0-beta.2 introduces breaking changes to the function version of .when as outlined in https://github.com/jquense/yup/pull/1542 . PR says change is breaking for function version, but changes are also breaking for the object version when `then` or `otherwise` are SchemaLikes instead of functions

**To Reproduce**

This test case fails on 1.0.0-beta.2 but works on 0.27.0
https://codesandbox.io/s/naughty-hermann-r0b2l?file=/src/index.test.js

**Expected behavior**
Either both test cases should work, or the docs should be updated. Currently the docs are unclear on the object version types, though the examples have been updated to use function for both `then` and `otherwise` However, this line: `You can provide an object literal where the key is is value or a matcher function, then provides the true schema and/or otherwise for the failure condition.` still implies that `then` can be a schema.

**Platform (please complete the following information):**
 Version 1.0.0-beta.2

jquense:
Yeah the docs may not have been fully updated or nots not super clear This change is intentional, only the function version is supported for branches

edgarsi:
I plead you to revisit this, to update CHANGELOG.md and README.md

Since 1.0.0 release is no longer beta, I was happily updating, and wasted time in confusion about this very issue. No doubt more people will grasp in the dark due to this.

jquense:
the readme is updated AFAIK happy to take PR's updating docs

PR #1917, opened 2023-02-15
Improve docs on migrating to 1.0.0

edgarsi:
Can you also edit [1906](https://github.com/jquense/yup/issues/1906) and add a link to [1579](https://github.com/jquense/yup/issues/1579) in the "Stricter when API" section?


jquense (approved):

jquense:
thanks!

Issue #1909, opened 2023-02-10
email is incorrectly validating


jiblett1000:
**Describe the bug**
As of the 1.0 release, email validation is now returning true when the value is something like:

`email@` // false 

`email@site` // true = incorrect

`email@site.` // false

`email@site.domain` // true

**To Reproduce**

```
const schema = object({
  email: string().email(),
});
```

**Expected behavior**

I think the above sums it up.

**Platform (please complete the following information):**
 - Browser: chrome
 - Version 1.0

**Additional context**

jquense:
I forgot about this change. Email validations follow: https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address to match what browsers use when validating email inputs. This was picked because there is an endless set of regexp's for emails each with their own false positives and negatives, and at least input validation is the one of the more common use-cases for yup.

We encourage you to override and change the validation to suit your needs.

jiblett1000:
@jquense Thanks for the response! After doing some reading, it makes total sense. My use case is just for a general email input, so I will leave it at the default I think.

pepijn-vanvlaanderen:
@jquense I disagree on this topic, I agree a regex can become very exhaustive to take all valid email address cases into account. 

However `email@site` is a case that should not be validated as a correct address.

I created a PR for this: https://github.com/jquense/yup/pull/1938

jiblett1000:
@pepijn-vanvlaanderen As hard as it is for me to accept it, I agree with @jquense on this one. Even though it is extremely uncommon to have an email such as `user@somesite`; it is possible. Considering an internal site, emails such as `user@localhost` can exist. 

But maybe adding a reasonable regex alternative in the docs would be an option? I know there can be a lot of variation there as well, but maybe providing the regex that was used previously by the `email()` validation method.

This thread provides some additional thoughts on the subject:

https://stackoverflow.com/questions/20573488/why-does-html5-form-validation-allow-emails-without-a-dot

pepijn-vanvlaanderen:
@jiblett1000 You are totally right! I forgot about those kinds of addresses.

Though I think in this case you should look to the practical side of it. But those kinds of email addressess are rarely used and probably not even validated at all with a validation library. So the trade-off here is to service an edge-case (which also can create their own override on the email regex) or the majority where the validation of `user@somesite` does not make sense.

For anyone stumbling upon this thread, here is the regex change I made in my PR:
```
[...]

PR #1938, opened 2023-03-08
Fix regex test for email addresses without tld

pepijn-vanvlaanderen:
The current regex in place does not properly validate when a TLD is omitted, for example this is valid: `email@site`
By making the last group in the regex required at least one time instead of zero, the issue is solved.

Fixes https://github.com/jquense/yup/issues/1909

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -19,7 +19,7 @@ import Schema from './schema';
 // Taken from HTML spec: https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
 let rEmail =
   // eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


ssanju01:
@pepijn-vanvlaanderen the current regex you are using in the PR marking email@site.e as valid, I feel there should be at least 2 characters after the dot 

Examples:
email@site.s - invalid 
email@site.si - valid

I tried the given regex, it seems to be working fine.
/^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/

Let me know your thoughts here!

jquense:
@pepijn-vanvlaanderen thanks for the PR, while I very much appreciate the contribution, I'm going to keep the current behavior, matching browsers

Issue #1891, opened 2023-01-27
Improvement: Tuple should be better represented in the describe method


DaddyWarbucks:
The `tuple` schema currently does not return any information about its inner types when using the `tuple([...]).describe()` method.

The schema currently maintains its inner types on a `types` property. This differs from the normal `array` schema that holds the state of its inner types on a property called `innerType`. I propose that we update the `tuple` schema to hold its state on the `innerType` as an array. And then we should export this in the describe method like an array schema.

See: https://github.com/jquense/yup/blob/master/src/tuple.ts

Renaming `types[]` to `innerTypes[]` on this schema is not required. But I'd like to make this information available in the describe method somehow. And I feel like just making `innerType` an array is the most familiar API. On an array the `innerType` will be a schema, on a tuple the `innerType` will be an array of schemas.

This is quite straight forward and I am happy to open a PR for it. But I wanted ask @jquense his opinion on it first.

PR #1947, opened 2023-03-16
Tuple describe() method

DaddyWarbucks:
This PR addresses: https://github.com/jquense/yup/issues/1891

The code aims to standardize how Array and Tuple inner types are configured. Historically `array` has stored it's one inner type on a prop `innerType`. Tuple has stored its types in `spec.types`. The types MUST be stored on the `spec` object for Tuple because it is needed for the `notType` method. See: https://github.com/jquense/yup/blob/f37a53f45048eefe445c14b27b1b031583f9585d/src/locale.ts#L137

Both Array and Tuple now have `schema.innerType` and `schema.spec.innerType`. This was done for backwards compatibility and consistency. And all corresponding methods and types have been updated.

This now allows the Tuple to return information about its types in the `describe` method

```js
const description = tuple([string(), number()]).describe();

description = {
  type: 'tuple',
  innerType: [
    { type: 'string },
    { type: 'number' }
  ]
}
```

Let me know if you have any questions!

### DIFF
--- a/src/locale.ts
+++ b/src/locale.ts
@@ -136,7 +136,7 @@ export let array: Required<ArrayLocale> = {
 export let tuple: Required<TupleLocale> = {
   notType: (params) => {
     const { path, value, spec } = params;
-    const typeLen = spec.innerType.length;
+    const typeLen = spec.types.length;
     if (Array.isArray(value)) {
       if (value.length < typeLen)
         return `${path} tuple value has too few items, expected a length of ${typeLen} but got ${
--- a/src/locale.ts
+++ b/src/locale.ts
@@ -136,7 +136,7 @@ export let array: Required<ArrayLocale> = {
 export let tuple: Required<TupleLocale> = {
   notType: (params) => {
     const { path, value, spec } = params;
-    const typeLen = spec.types.length;
+    const typeLen = spec.innerType.length;
     if (Array.isArray(value)) {
       if (value.length < typeLen)
         return `${path} tuple value has too few items, expected a length of ${typeLen} but got ${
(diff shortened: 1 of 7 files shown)


src/locale.ts
```diff
@@ -136,7 +136,7 @@ export let array: Required<ArrayLocale> = {
 export let tuple: Required<TupleLocale> = {
   notType: (params) => {
     const { path, value, spec } = params;
-    const typeLen = spec.types.length;
+    const typeLen = spec.innerType.length;
```
jquense, line 139: i'd rather keep the `types` name, i wanted to migrate array to use it it as well but opted to avoid the breaking change

DaddyWarbucks: OK. I can change that back to `types`. I suppose that is actually a breaking change too, so it should remain `types`. Its a bummer I didn't get to this before the official release of V1 🤷 

What should I do with Array?
1 - Return it to its current shape (no spec)
2 - Add `spec.innerType = {...the one type}`
3 - Add `spec.types = [{ ...the one type }]` // note the array
4 - Add `spec.types = { ...the one type }` // note no array
5 - Add `spec.type = { ...the one type }` // note spec is singular

I am leaning towards #3. It seems most consistent with Tuple. And if there is ever another use case for some "inner type", then the property `types` with an array is flexible.

DaddyWarbucks:
While I've got your eyes on this, this appears to be a bug in the array schema: https://github.com/jquense/yup/blob/b0e67ef173dbb04b1fbf0eb163b49d3bde1f3d6a/src/array.ts#L272
Shouldn't that pass `innerOptions`?

src/locale.ts
jquense: I think 3 makes sense to me 👍

DaddyWarbucks:
@jquense All updates have been made. let me know if you have anymore questions or concerns.

jquense:
thanks!

PR #1971, opened 2023-04-05
Respect context for object's children

FedorSherbakov:
Pass context to nested elements for both `getDefault` and `cast` That solves the issue where conditional defaults of nested elements were not resolving correctly if they had context dependency

Added tests are covering examples of use cases that are being fixed

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -156,7 +156,7 @@ export default class ObjectSchema<
     let value = super._cast(_value, options);
 
     //should ignore nulls here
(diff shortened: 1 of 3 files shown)


jquense (approved):

jquense:
Thanks!

FedorSherbakov:
🙌 when do you think this could be published?

jquense:
out in 1`.1.0`

FedorSherbakov:
Thank you!

PR #1972, opened 2023-04-06
Add currentValue to schema

webblocksapp:
**New Feature**

When using yup with libraries like Formik or React Hook Forms in React, the entire yup schema is revalidated whenever an input is filled. This becomes problematic when server-side validation needs to be added, as every time a form input is typed, the whole schema is validated, triggering an API request. For instance, typing in the password field might cause the test function for the email field to run. 

To solve this issue, consider adding the current value to the schema for comparing it with the new value. If they are the same, the test function is not run, resulting in more efficient validation.

This is not a yup issue, only a new feature to integrate with existing javascript form validation libraries that use yup and require more control for server-side validations.

**Schema example:**

```ts
const validationSchema = yup.object({
  email: yup
    .string()
    .email()
    .required('Email is required')
    .test('email-exists', 'Email Validation Error', (value, ctx) => {
      if (value !== ctx.currentValue) { //<=== New feature added ctx.currentValue
        return new Promise(async (resolve) => {
          const response = await axios.get(`http://localhost:3000/users?email=${value}`);
          resolve(response.data.length ? ctx.createError({ message: 'Email Exists' }) : true);
        });
      }
    }),
  password: yup.string().min(8, 'Password should be of minimum 8 characters length').required('Password is required'),
});
```
`ctx.currentValue` refers to the current value the field has in the form, which will be compared with `value`, which is the new user input. `ctx.currentValue` is different than `ctx.originalValue`, this last one is the non-casted `value`.

**Steps for testing this current branch:**

1. Download the following example repository:
https://github.com/webblocksapp/react-formik-yup-server-validation

2. Compile the yup branch.
3. Inside the downloaded repo, run `pnpm link <yup-repository-dir>`
4. Run `pnpm dev` and `pnpm dev-server` to visualize the form.
5. Type the password field, you will see the console logs running for the email field.

**Expected behavior**

You will see on the repo example that the `.test` function content for the `email` field, will run once
`console.log('Email field test logic is executed')` when the password field is input.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,11976 @@
+lockfileVersion: '6.0'
+
+dependencies:
+  property-expr:
+    specifier: ^2.0.5
+    version: 2.0.5
+  tiny-case:
+    specifier: ^1.0.3
+    version: 1.0.3
+  toposort:
+    specifier: ^2.0.2
+    version: 2.0.2
+  type-fest:
+    specifier: ^2.19.0
+    version: 2.19.0
+
+devDependencies:
+  '@4c/cli':
+    specifier: ^4.0.4
+    version: 4.0.4(eslint@8.34.0)(typescript@4.9.5)(webpack@5.78.0)
+  '@4c/rollout':
+    specifier: ^4.0.2
+    version: 4.0.2
+  '@4c/tsconfig':
+    specifier: ^0.4.1
+    version: 0.4.1(typescript@4.9.5)
+  '@babel/cli':
+    specifier: ^7.21.0
+    version: 7.21.0(@babel/core@7.21.0)
+  '@babel/core':
+    specifier: ^7.21.0
+    version: 7.21.0
+  '@babel/preset-typescript':
+    specifier: ^7.21.0
+    version: 7.21.0(@babel/core@7.21.0)
+  '@rollup/plugin-babel':
+    specifier: ^5.3.1
+    version: 5.3.1(@babel/core@7.21.0)(rollup@2.79.1)
+  '@rollup/plugin-node-resolve':
+    specifier: ^13.3.0
+    version: 13.3.0(rollup@2.79.1)
+  '@types/jest':
+    specifier: ^27.5.2
+    version: 27.5.2
+  '@typescript-eslint/eslint-plugin':
+    specifier: ^5.53.0
+    version: 5.53.0(@typescript-eslint/parser@5.53.0)(eslint@8.34.0)(typescript@4.9.5)
+  '@typescript-eslint/parser':
+    specifier: ^5.53.0
+    version: 5.53.0(eslint@8.34.0)(typescript@4.9.5)
+  babel-jest:
+    specifier: ^27.5.1
+    version: 27.5.1(@babel/core@7.21.0)
+  babel-preset-env-modules:
+    specifier: ^1.0.1
+    version: 1.0.1
+  doctoc:
+    specifier: ^2.2.1
+    version: 2.2.1
+  dts-bundle-generator:
+    specifier: ^6.13.0
+    version: 6.13.0
+  eslint:
+    specifier: ^8.34.0
+    version: 8.34.0
+  eslint-config-jason:
+    specifier: ^8.2.2
+    version: 8.2.2(@babel/core@7.21.0)(@typescript-eslint/eslint-plugin@5.53.0)(eslint-plugin-import@2.27.5)(eslint-plugin-react-hooks@4.6.0)(eslint-plugin-react@7.32.2)(eslint@8.34.0)(typescript@4.9.5)
+  eslint-config-prettier:
+    specifier: ^8.6.0
+    version: 8.6.0(eslint@8.34.0)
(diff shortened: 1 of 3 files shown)


jquense:
Hey there, thanks for PR. I don't think this is the best way to solve this. Yup schema are immutable by design and a lot of code assumes that, I don't want to change and make schema stateful now.

Ultimately form libraries shouldn't re run validation for entire schema if only one field changes, that's up to them to optimize (and they should, formik at least does this already) if you want to store some value to use in the test to decide to rerun there is already a `context` value that can be passed in for this purpose

webblocksapp:
Hi @jquense , thanks for your soon reply. 

I agree with you that form libraries should handle granular validation, but unfortunately, most of them are not designed over this principle which will be extremely useful for server-side validation. Since running all frontend validations is not expensive, granular validations are not often applied.

Implementing granular validation in existing form libraries as an open-source contribution is not easy since it would require a breaking change in their form engine and extensive testing. Granular validation requires a tree search algorithm to validate branches of the data tree, which is currently not supported by existing form libraries. It is easier to re-validate the entire schema, especially for conditional rendering.

Another issue with form libraries like `react-hook-form` is that they provide adapters to work with zod, joi, and other data assertion libraries. However, the issue with those others is that not all provide a validateAt method, limiting the implementation of granular validations.

I've been studying the possibility to work with `validateField` from formik but won't be enough, I would need to build a framework upon formik for covering most complex scenarios like conditional rendering and nested validations.

So I think this solution of storing the last validated value inside the `yup` schema would be extremely powerful, so you can have more control when creating custom validations or using the `.test` function for creating a server side validation when working with form validation libraries. As I mentioned validations on the frontend side are not expensive, the only expensive ones are server-side which can be controlled by comparing the current value vs the previous validated value.

I was thinking, to being more consistent with your library, to store the `currentValue` inside the schema `spec` object (at the same level where the default value is stored), and maybe simply call it `value` or `lastValidatedValue`.

```javascript
{
    "strip": false,
    "strict": false,
    "abortEarly": true,
    "recursive": true,
    "nullable": false,
    "optional": false,
    "coerce": true,
    "default": "test@mail.com",
    "value":  "current@mail.com",
}
```

Please let me know your thoughts on this approach or if you have a cleaner solution that I could implement.

jquense:
Formik already does granular validation via `validateAt`. However your goal seems to be to prevent multiple extra HTTP calls when doing server side validation. If that's right, neither yup or the form library is the correct place for this. This should be handled by code making the HTTP request (lots of api clients have this built in). I would suggest caching the results of yor validation calls or memoizing them based on the inputs:

```tsx
const validateEmail = memoizeOn((email) =>  axios.get(`http://localhost:3000/users?email=${email}`))

const validationSchema = yup.object({
  email: yup
    .string()
    .email()
    .required('Email is required')
    .test('email-exists', 'Email Validation Error', async (value, ctx) => {
        const response = await validateEmail(value);
        resolve(response.data.length ? ctx.createError({ message: 'Email Exists' }) : true);
    }),
  password: yup.string().min(8, 'Password should be of minimum 8 characters length').required('Password is required'),
});
```

Or something similar. But, storing the current value on the schema itself **will not work** because yup schema are not stateful. Something like this would break if you used a schema in multiple forms for instance.

webblocksapp:
Thank you for providing the solution. However, the memoized validator is still stateful. To avoid this, you can wrap it in a function that generates a new instance of `validateEmail` every time it's needed:

```
const validateEmailValidator = () => memoizeOne((email) =>  xios.get(`http://localhost:3000/users?email=${email}`))
const validateEmail = validateEmailValidator()
```

The same can be achieved by creating a schema instance as follows:
```
const schema = () => yup.object(...)
const schemaInstance = schema()
```

This would allow the schema to store the current value for handling caching and serve multiple forms that require the same schema.

On the other hand, the validateAt method is not sufficient because it only provides recursive validation from parent to child, but not branch validation from child to parent for nested schemas. Consider the following scenario:

```typescript
const schema = yup.object({
  contact: yup.object({
     name: yup.string()
  })
});

const data = { contact: { name: 'Foo' }};
```

When schema.validateAt('contact.name', data) is executed, only SCHEMA.contact.name will be marked as valid, but the parent branch SCHEMA.contact and the root SCHEMA will not be validated. Therefore, a framework is needed over the form library to validate the host branch of the name string schema. So for this reason is easier to re-run the whole schema, and would be more powerfull if the schema brings the current value for manually caching it.

```typescript
      if (value !== ctx.currentValue) { //<=== Manual caching
        return new Promise(async (resolve) => {
          const response = await axios.get(`http://localhost:3000/users?email=${value}`);
          resolve(response.data.length ? ctx.createError({ message: 'Email Exists' }) : true);
        });
      }
```

jquense:
Appreciate your taking the time to put this together and thinking through the problem a bunch. Ultimately I don't think it makes sense to  do this in yup as there are a number of less error prone ways to achieve efficient revalidation of slow tests.

I would suggest that you may be a little to focused on trying to prevent multiple validation runs. In my experience using yup with form validation it's better to make the tests faster and cache results, or denounce http calls instead of trying to prevent re-runs.

Issue #1952, opened 2023-03-17
Nullable date should not parse as invalid date in v1.0.2


exah:
**Describe the bug**

`Yup.date().nullable().required('Required')` produces invalid date error when initial value is `null`:

```
key must be a `date` type, but the final value was: `Invalid Date`
```

**To Reproduce**

https://codesandbox.io/s/cool-swartz-b4pfbp?file=/src/index.test.js

**Expected behavior**

Should return `ValidationError` with `'Required'` message.

**Platform (please complete the following information):**

 - Firefox 
 - 111

**Additional context**

Stopped working in v1.0.2.

Temporary solution is to replace `yup.date()` with `yup.mixed((val): val is Date => val instanceof Date)`, but `min`, `max` methods needs to be re-implemented as well.

jquense:
its not nullable in your example, you `Yup.date().nullable().required('Required')` the required overrides the `nullable`, this is a change in v1, see https://github.com/jquense/yup/issues/1906

exah:
Hey @jquense, thanks for the reply.

I understood that. I need to remove `.nullable()` as `.required()` already checks it for me, but I have one follow-up question. 

Even if I remove `nullable` and use just `required()`, I get the same invalid date error: 

https://codesandbox.io/s/bold-zeh-2btd0d?file=/src/index.test.js

Although if I replace `.date()` with `.mixed()`, everything works as expected because there is no parsing happens during the validation (expected behaviour for me): 

https://codesandbox.io/s/jovial-phoebe-63sidy?file=/src/index.test.js

Should I switch to `.mixed` in my case?

bel0v:
@jquense hi, the issue stays with `Yup.date().required('Required')`

https://codesandbox.io/s/weathered-worker-xw3tov?file=/src/index.test.js:100-131

so it seems to be a bug related to `date()` casting, not nullable override you mentioned

jquense:
it works if the date is actually nullable `date().nullable().cast(null)`

jarnaiz:
Hello,

I dont understand the new API:

```js
it("required string pass", async () => {
  await expect(
    yup.string().required("Required").validate(null)
  ).rejects.toEqual(new yup.ValidationError("Required"));
});

it("required date fails", async () => {
  await expect(
    yup.date().required("Required").validate(null)
  ).rejects.toEqual(new yup.ValidationError("Required"));
});

it("required date pass", async () => {
  await expect(
    yup.date().required().typeError("Required").validate(null)
[...]

PR #1982, opened 2023-04-14
Potentially breaking: Make null validation errors consistent across schema

jquense:
Previously, schema that parse to invalid values like `InvalidDate` or `NaN` always tripped the `typeError` validation before the nullability error. This was the due to typeError no longer firing on `null` values, with the expectation that it would always hit the newer nullability check. 

This change fixes the issue for Number and Date schema, so that they act the same as String and other schema, where `null` values do not trigger the typeError validation. The consequence of this is that casting `null` values for non-nullable number schema, now return `null` instead of `NaN` and dates, return `null` instead of `InvalidDate`. Essentially these schema will not fruitlessly attempt to coerce `null` into a number or date anymore. 

I am marking this as a bug fix since it is, and as with most bug fixes, could also be considered a breaking change. I'm going to optimistically not cut a v2 for this. If you have unexpected changes it is likely due to number/date schema hitting different validation errors (nullability). 

Previously for these schema it was common practice to change the `typeError` for these schema to something like `"this is required"`assuming the issue was due to invalid `null`s. NOW the schema will show the `locale.notNull` message, so if you may want to adjust that to something more user friendly if you surface default locale messages to users (we don't generally recommend that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ const num = number().cast('1'); // 1
 const obj = object({
   firstName: string().lowercase().trim(),
 })
+  .json()
   .camelCase()
   .cast('{"first_name": "jAnE "}'); // { firstName: 'jane' }
 ```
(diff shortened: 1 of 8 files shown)


jquense:
fixes @jarnaiz issue in #1952

Issue #2002, opened 2023-05-12
[request] export printValue function from the bundle


diegoulloao:
**Describe the bug**
I have a yup "es" translation library that uses the `printValue` function from yup, it used to work in older versions where this function was exported but it doesn't anymore.

Some projects are depending of this translation library I've created, they can't update to the latest yup version.

This `printValue` function has other dependencies like `printSimpleValue` and I don't think is a good idea to just copy and paste those function definitions into my code in order to do the fix.

I'm requiring from yup to export this `printValue` function in the bundle in order to allow translation packages that uses this function to work properly.

If you want to know more about this package (which is very simple), please follow this link: https://github.com/diegoulloao/yup-es

This is the related issue: https://github.com/diegoulloao/yup-es/issues/7

It is possible from you to consider this petition?

Thank you for your time.

jquense:
Happy to expose this, want to send a PR?

diegoulloao:
Hey @jquense, just coming back to this, I was tied up in my work.
I can see you pushed a commit that exposes this. What's the release status?

diegoulloao:
Ok seems like fixed in this version: https://github.com/jquense/yup/releases/tag/v1.2.0
Thank you!

PR #2008, opened 2023-05-23
feat: expose printValue (#2002)

dilanluna:
Export this `printValue` function in the bundle in order to allow translation packages that uses this function to work properly.


jquense (approved):

jquense:
Thanks!

Issue #2027, opened 2023-06-13
Is ObjectSchema's class constructor recommended?


justinbhopper:
While migrating to v1, I noticed that the schemas are now exposed as a Class (not just an interface).

This allows me to use `new ObjectSchema<T>(...)`, instead of the traditional `object(...)`.

The key benefits I can see here are:
1. Intellisense recommends the properties that can (and need to) be defined on the schema.
2. TypeScript is narrows down which property(s) don't match the target type, rather than just saying the entire schema is wrong.

See a comparison here:
![image](https://github.com/jquense/yup/assets/29075873/8ff4b5d6-b8b3-4d48-87ef-927f61ddbb24)

### So my question:
Is this recommended practice?  The examples/documentation doesn't make mention of it, so I'm afraid to utilize it if its not part of the official API.

justinbhopper:
I've added a PR that updates the `object()` schema to allow explicitly defining the target type.  This gives us the same benefits as my example above, but not having to resort to using the ObjectSchema class constructor.

PR #2028, opened 2023-06-13
Allow explicit type to be defined for object()

justinbhopper:
Relates to #2027

This change would allow us to explicit define the target type for `object()` ahead of time via the first generic arg, allowing for more direct intellisense feedback in TypeScript.

Example:
image

Please note, this change is fully backwards compatible.  The generic arg is optional and the output type is still inferred properly if the not provided.

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -79,10 +79,17 @@ function unknown(ctx: ObjectSchema<any, any, any>, value: any) {
 
 const defaultSort = sortByKeyOrder([]);
 
-export function create<TIn extends AnyObject, C extends Maybe<AnyObject> = AnyObject, S extends Shape<TIn, C> = Shape<TIn, C>>(spec: S): ObjectSchema<_<TypeFromShape<S, C>>, C, _<DefaultFromShape<S>>>;
-export function create(): ObjectSchema<{}, AnyObject, _<DefaultFromShape<{}>>>;
-export function create<TIn extends AnyObject, C extends Maybe<AnyObject> = AnyObject>(spec?: Shape<TIn, C>) {
-  type TDefault = _<DefaultFromShape<TIn>>;
+export function create(): ObjectSchema<{}, AnyObject, {}>;
+
+export function create<
+  TIn extends AnyObject, 
+  C extends Maybe<AnyObject> = AnyObject, 
+  S extends Shape<TIn, C> = Shape<TIn, C>
+>(spec: S): ObjectSchema<_<TypeFromShape<S, C>>, C, _<DefaultFromShape<S>>>;
+
+export function create<C extends Maybe<AnyObject> = AnyObject, S extends ObjectShape = {}>(spec?: S) {
+  type TIn = _<TypeFromShape<S, C>>;
+  type TDefault = _<DefaultFromShape<S>>;
 
   return new ObjectSchema<TIn, C, TDefault>(spec as any);
 }
--- a/src/object.ts
+++ b/src/object.ts
@@ -79,12 +79,10 @@ function unknown(ctx: ObjectSchema<any, any, any>, value: any) {
 
 const defaultSort = sortByKeyOrder([]);
 
-export function create<
-  C extends Maybe<AnyObject> = AnyObject,
-  S extends ObjectShape = {},
->(spec?: S) {
-  type TIn = _<TypeFromShape<S, C>>;
-  type TDefault = _<DefaultFromShape<S>>;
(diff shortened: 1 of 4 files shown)


jquense:
What close? Seemed like a nice improvement

justinbhopper:
@jquense Apologies -- I ran into some test cases where this change seemed to break.  I've made some further progress fixing those issues, but there is still one issue involving optional properties (e.g. `{ foo?: string }`).  

I had planned on re-opening the PR if I had success, but since you're open to the idea I'll keep it as draft instead.  I've updated the PR that illustrates the issue I'm currently stuck on.  

image

justinbhopper:
Made progress fixing the errors, but it required me to drop the use of `MakeKeysOptional`.  I don't understand its true purpose, so I'd like your feedback.  Perhaps you can look to see if it breaks any expectations you had, and if so we can add some tests in `test/types.ts` to cover those scenarios?

justinbhopper:
@jquense I've fixed the test cases -- marking this as ready to review again

justinbhopper:
@jquense A nice added bonus to this PR is the previous way of defining _also_ gets the same benefits

Before:
image

After:
image

src/object.ts
```diff
@@ -79,10 +78,36 @@ function unknown(ctx: ObjectSchema<any, any, any>, value: any) {
...
+  : Target;
+
+export function create(): ObjectSchema<{}, AnyObject, {}>;
+
 export function create<
+  TIn extends AnyObject,
   C extends Maybe<AnyObject> = AnyObject,
-  S extends ObjectShape = {},
->(spec?: S) {
+  S extends MatchShape<TIn, C> = MatchShape<TIn, C>
```
jquense, line 107: A challenge here is that it's no longer possible to specify the context without explicitly defining TIn, e.g  `create<Context>({ ... })`

src/object.ts
```diff
@@ -127,7 +152,7 @@ export default class ObjectSchema<
   TContext = AnyObject,
   TDefault = any,
   TFlags extends Flags = '',
-> extends Schema<MakeKeysOptional<TIn>, TContext, TDefault, TFlags> {
+> extends Schema<TIn, TContext, TDefault, TFlags> {
```
jquense, line 155: is it safe to remove the optional key util here?

src/object.ts
```diff
@@ -374,7 +399,7 @@ export default class ObjectSchema<
     additions: U,
     excludes: readonly [string, string][] = [],
   ) {
-    type UIn = TypeFromShape<U, TContext>;
+    type UIn = MakePartial<TypeFromShape<U, TContext>>;
```
jquense, line 402: what's happening here?

src/object.ts
justinbhopper: Yes, we don't want to mislead what the `__outputType` actual result is going to be.  Its up to the constructor (or in our case, the `object` function) to force optional keys upon TIn where necessary.  This behavior was moved into the `TypeFromTarget` via the `MakePartial` type.

src/object.ts
justinbhopper: Because `MakeKeysOptional` is no longer forced upon all `ObjectSchema` via the `extends`, this means `shape` needs to perform this same operation instead so it retains the same behaviors as before.

Essentially I lifted `MakeKeysOptional` up to the factories and the `shape` method.

src/object.ts
justinbhopper: Is `Context` ever explicitly defined via `object()`?  I thought it was only passed down/inherited from other methods.  I couldn't find any examples of `object<...>` in the codebase or documentation, so I wasn't aware this was something supported.

Do you have an example to illustrate?

jquense: Context is always explicitly provided if used. It defines the arbitrary extra data made available to tests and transforms. E.g. 

```
object<{ api: APIClient }>().test ((value, schema)=> schema.options.context.api)
```

justinbhopper: Okay, I see now.  Its defined ahead of time, but the value is passed in during validation calls via `ValidateOptions`.

I think unfortunately this means fundamentally the first generic arg can't be the schema's type.

What are your thoughts if I move the explicit type ability to `.shape()` method instead?

justinbhopper: @jquense Did you see my question above?

jquense:
Sorry I haven't had time to look at this! When I do have time I hope to look. I can reopen it then if I get to it. Thanks again for taking the time to put this together

PR #2030, opened 2023-06-15
[FIX] fix string email validation

DafaZakhulhaq27:
previously if there was entering "test@gmailcom" it passed even though there was no ".". Now I add validation if there is no "." can't be passed


jquense:
Thanks for the PR but the current behavior is intentional, to match browser email validation. See the v1 migration and docs for more context. You can extend our override the method locally if you want more specific beavhior

Issue #2034, opened 2023-06-21
Email 'jhon@smith' pass as valid


margori:
**Describe the bug**
Email 'jhon@smith' is passing as valid email.

**To Reproduce**

```
const yup = require("yup");

it("should fail", async () => {
  const schema = yup.object({
    email: yup.string().email()
  });

  const obj = {
    email: "jhon@smith"
  };

  await expect(schema.isValid(obj)).resolves.toBe(false);
});
```

**Expected behavior**

Email 'jhon@smith' should not be valid.

jquense:
https://github.com/jquense/yup#stringemailmessage-string--function-schema


Email uses the same regex as browsers, which is intentional and yes, limited.

PR #2038, opened 2023-06-25
#2034 Fix Email Regex

Vikas208:
#2034 
Fix Email Address Regex fails for the input which doesn't have the domain name.


jquense:
Thanks for the PR, we aren't accepting changes to the email regex, it's intentionally aligned with browser email input validation

Issue #2043, opened 2023-07-02
Slow performance with large arrays


tedeschia:
- Write a title that summarizes the specific problem
- Describe what you are trying to accomplish AND what you have tried
I need to validate some large arrays (about 100,000 items). When there are lots of validation errors, the validation process takes a long time to run.

I reproduced the issue in CodeSandbox: https://codesandbox.io/s/yup-performance-test-6jxn5m?file=/src/index.js

There I defined an array schema of required numbers and run the validation for some datasets to measure the performance. This is the result:

Validation of 5000 items took 169.6 ms 
Validation of 10000 items took 320.1 ms - 88.7% increase vs last fun
Validation of 20000 items took 985.1 ms - 207.7% increase vs last fun
Validation of 40000 items took 5,034.8 ms - 411.1% increase vs last fun

It seems to be an exponential increase of time to process the validations.

Doing some profiling, it seems that the issue is in this line of ValidationError class:
`this.inner = this.inner.concat(err.inner.length ? err.inner : err);`

jquense:
That would probably be pretty easy to optimize if you want to take a stab at it?

jquense:
Alternatively, you could at abortEarly: true to fail fast, since it's unlikely you need 40,000 of probably the same error

tedeschia:
abortEarly is not an option for me. It's nos necessarily all the same errors (this case was just an example) and I need to give the user a feedback for every row.

I tried with this change and it was solved:

```
const innerErrors = err.inner.length ? err.inner : [err];
this.inner.splice(this.inner.length, 0, ...innerErrors);
```
do you want me to send a PR?

I'm also measuring other bottlenecks and found that ValidationError constructor also takes its time (considering thousands of it) at:
`super()`
`Error.captureStackTrace(this, ValidationError)`
I'm wondering if I could implement any kind of "minimalistic" version of ValidationError class, in case of people (like me) that prioritize performance instead of functionallity (I'm just interested in error message, and don't use stack trace at all).

jquense:
I'd definitely be open to a PR for both, maybe we could toggle the stack traces off via an option, tbh the stack isn't really useful for validation error

tedeschia:
With the proposed optimizations in #2044, the performance is significantly improved, and there is no exponential performance degradation when the array grows.
But anyway, it takes some time to perform the validations. In my tests, an array of 60000 items takes about 4 seconds to get validated, and during this process, the UI gets blocked.
[...]

PR #2044, opened 2023-07-03
fix: performance improvement (#2043)

tedeschia:
Resolves performance issues for large data validation described on #2043 :
- Arrays of ValidationError
  - Mutate the array with splice instead of using concat
- ValidationError class
  - Add the new option "disableStackTrace" to stop capturing stack trace when instantiating ValidationError
  - Stop ValidationError extending Error, so no need to call super() when instantiating ValidationError

### DIFF
--- a/src/ValidationError.ts
+++ b/src/ValidationError.ts
@@ -55,7 +55,7 @@ export default class ValidationError implements Error {
       if (ValidationError.isError(err)) {
         this.errors.push(...err.errors);
         const innerErrors = err.inner.length ? err.inner : [err];
-        this.inner.splice(this.inner.length, 0, ...innerErrors);
+        this.inner.push(...innerErrors);
       } else {
         this.errors.push(err);
       }
--- a/src/ValidationError.ts
+++ b/src/ValidationError.ts
@@ -5,7 +5,10 @@ let strReg = /\$\{\s*(\w+)\s*\}/g;
 
 type Params = Record<string, unknown>;
 
-export default class ValidationError extends Error {
+export default class ValidationError implements Error {
+  name: string;
+  message: string;
+  stack?: string | undefined;
   value: any;
   path?: string;
   type?: string;
@@ -38,9 +41,8 @@ export default class ValidationError extends Error {
     value?: any,
     field?: string,
     type?: string,
+    disableStack?: boolean,
   ) {
-    super();
-
     this.name = 'ValidationError';
     this.value = value;
     this.path = field;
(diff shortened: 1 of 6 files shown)


src/schema.ts
```diff
@@ -499,7 +503,11 @@ export default abstract class Schema<
 
       test(args!, panicOnce, function finishTestRun(err) {
         if (err) {
-          nestedErrors = nestedErrors.concat(err);
+          nestedErrors.splice(
+            nestedErrors.length,
+            0,
+            ...(Symbol.iterator in err ? err : [err]),
+          );
```
jquense, line 510:
```suggestion
          Array.isArray(err) ? nestedErrors.push(...err) :  nestedErrors.push(err)
```

src/ValidationError.ts
```diff
@@ -52,7 +54,8 @@ export default class ValidationError extends Error {
     toArray(errorOrErrors).forEach((err) => {
       if (ValidationError.isError(err)) {
         this.errors.push(...err.errors);
-        this.inner = this.inner.concat(err.inner.length ? err.inner : err);
+        const innerErrors = err.inner.length ? err.inner : [err];
+        this.inner.splice(this.inner.length, 0, ...innerErrors);
```
jquense, line 58:
```suggestion
        this.inner.push(...innerErrors);
```

tedeschia:
@jquense I think I made all requested changes, is there anything pending from me to approve the PR?

jquense (approved):

jquense:
thanks!

jquense:
@tedeschia unfortunately this broke a lot of tests so i had to revert. If you want re-open and fix the failures i'll happily merge it again

tedeschia:
yes I'll do it! and will submit the changes required to fix the failing tests as soon as possible.
thanks

________________________________
De: Jason Quense ***@***.***>
Enviado: viernes, 28 de julio de 2023 10:14
Para: jquense/yup ***@***.***>
Cc: Adrian Tedeschi ***@***.***>; Mention ***@***.***>
Asunto: Re: [jquense/yup] fix: performance improvement (#2043) (PR #2044)


@tedeschia<https://github.com/tedeschia> unfortunately this broke a lot of tests so i had to revert. If you want re-open and fix the failures i'll happily merge it again

tedeschia:
@jquense the reason of tests failing is having changed the inheritance of ValidationError from Error. 
I broke that inheritance because thowsands of "new Error()" was one of the main time consuming tasks in my profiling, and changed by an implementation of Error interface.

But that change make this kind of assertion to fail:
".rejects.toThrowError()"

Reverting "ValidationError implements Error" to "ValidationError extends Error" makes all the tests run OK.

Do you think I can keep the ValidationError change and rewrite the tests to handle this new kind of ValidationError object? If you are OK I can work on it.

Thanks!

tedeschia:
@jquense yes you are right. inspecting jest toThrow code I noticed it checks for error object this way:
```
switch (Object.prototype.toString.call(value)) {
    case '[object Error]':
    case '[object Exception]':
    case '[object DOMException]':
      return true;
    default:
      return value instanceof Error;
  }
```

so adding `[Symbol.toStringTag] = 'Error'` to ValidationError class did the trick.
I have created a new PR #2072 with this fix.

Issue #2050, opened 2023-07-06
ObjectSchema describe doesn't resolve conditions


SpiderDan98:
**Describe the bug**
If you call `describe` on an ObjectSchema with a `when` function, the `fields` of the ObjectSchema won't be extended by the fields of the when condition.

**To Reproduce**

[codesandbox](https://codesandbox.io/s/jovial-scooby-z9p37c?file=/src/index.test.js)

**Expected behavior**
The [SchemaDescription](https://github.com/jquense/yup#schemadescribeoptions-resolveoptions-schemadescription) should be include all fields within the conditional fields.

**Platform:**
 - Browser chrome
 - Version 113.0.5672.92 (arm64)

**Additional context**
The Schema `describe` works very well and resolve conditional tests 👍

DaddyWarbucks:
We should also double check this on the other schemas that have children like `array` and `tuple`

jquense:
That looks right, PR?

DaddyWarbucks:
I'll see what I can do later this evening or tomorrow.

lukefredrickson:
@DaddyWarbucks @jquense Has there been any more progress on this issue?

DaddyWarbucks:
There has been some discussion over on the PR. I think we know the problem but are discussing the best solution. I will try to get back to it next week.

PR #2051, opened 2023-07-08
Allow parent resolve before describe.

DaddyWarbucks:
Solves: https://github.com/jquense/yup/issues/2050

This PR updates the `object`, `array`, and `tuple` schemas such that they are resolved before being described. This ensures that the `describe` method works for things like `object({...}).when(...)`

### DIFF
--- a/src/array.ts
+++ b/src/array.ts
@@ -268,7 +268,7 @@ export default class ArraySchema<
     );
   }
 
-  describe(options?: ResolveOptions<TContext>) {
+  protected _describe(options?: ResolveOptions<TContext>) {
     let base = super.describe(options) as SchemaInnerTypeDescription;
     if (this.innerType) {
       let innerOptions = options;
@@ -282,6 +282,11 @@ export default class ArraySchema<
       base.innerType = this.innerType.describe(innerOptions);
     }
     return base;
+  }
+
+  describe(options?: ResolveOptions<TContext>) {
+    const next = options ? this.resolve(options) : this;
+    return next._describe(options)
   }
 }
 
--- a/src/array.ts
+++ b/src/array.ts
@@ -268,9 +268,10 @@ export default class ArraySchema<
     );
   }
 
-  protected _describe(options?: ResolveOptions<TContext>) {
-    let base = super.describe(options) as SchemaInnerTypeDescription;
-    if (this.innerType) {
+  describe(options?: ResolveOptions<TContext>) {
+    const next = (options ? this.resolve(options) : this).clone();
+    const base = super.describe(options) as SchemaInnerTypeDescription;
+    if (next.innerType) {
       let innerOptions = options;
       if (innerOptions?.value) {
         innerOptions = {
@@ -279,14 +280,9 @@ export default class ArraySchema<
           value: innerOptions.value[0],
         };
       }
-      base.innerType = this.innerType.describe(innerOptions);
+      base.innerType = next.innerType.describe(innerOptions);
     }
     return base;
-  }
-
-  describe(options?: ResolveOptions<TContext>) {
-    const next = options ? this.resolve(options) : this;
-    return next._describe(options)
   }
 }
 
--- a/src/object.ts
+++ b/src/object.ts
@@ -507,10 +507,11 @@ export default class ObjectSchema<
     return this.transformKeys((key) => snakeCase(key).toUpperCase());
   }
 
-  protected _describe(options?: ResolveOptions<TContext>) {
-    let base = super.describe(options) as SchemaObjectDescription;
+  describe(options?: ResolveOptions<TContext>) {
+    const next = (options ? this.resolve(options) : this).clone();
+    const base = super.describe(options) as SchemaObjectDescription;
     base.fields = {};
-    for (const [key, value] of Object.entries(this.fields)) {
+    for (const [key, value] of Object.entries(next.fields)) {
       let innerOptions = options;
       if (innerOptions?.value) {
         innerOptions = {
@@ -522,11 +523,6 @@ export default class ObjectSchema<
       base.fields[key] = value.describe(innerOptions);
     }
     return base;
-  }
-
-  describe(options?: ResolveOptions<TContext>) {
-    const next = options ? this.resolve(options) : this;
-    return next._describe(options)
   }
 }
 
--- a/src/object.ts
+++ b/src/object.ts
@@ -507,7 +507,7 @@ export default class ObjectSchema<
     return this.transformKeys((key) => snakeCase(key).toUpperCase());
   }
 
-  describe(options?: ResolveOptions<TContext>) {
+  protected _describe(options?: ResolveOptions<TContext>) {
     let base = super.describe(options) as SchemaObjectDescription;
     base.fields = {};
     for (const [key, value] of Object.entries(this.fields)) {
@@ -522,6 +522,11 @@ export default class ObjectSchema<
       base.fields[key] = value.describe(innerOptions);
     }
     return base;
+  }
+
+  describe(options?: ResolveOptions<TContext>) {
+    const next = options ? this.resolve(options) : this;
+    return next._describe(options)
   }
 }
 
--- a/src/tuple.ts
+++ b/src/tuple.ts
@@ -159,9 +159,10 @@ export default class TupleSchema<
     });
   }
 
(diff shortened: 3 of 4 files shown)


SpiderDan98 (approved):
Thanks for the pretty quick fix

src/array.ts
```diff
@@ -283,6 +283,11 @@ export default class ArraySchema<
     }
     return base;
   }
+
+  describe(options?: ResolveOptions<TContext>) {
```
jquense, line 287: can we combine the two here, and just keep the one override?

DaddyWarbucks: Perhaps, but I struggled to understand how. I started off with something like
```ts
 describe(options?: ResolveOptions<TContext>) {
    const next = options ? this.resolve(options) : this.clone();
    let base = next.describe(options) as SchemaObjectDescription;
    ...
  }
``` 
But the problem is the `.resolve()` and `.clone()` both return a new schema. So that following `next.describe()` causes an infinite loop.

DaddyWarbucks: I could do this
```ts
 describe(options?: ResolveOptions<TContext>) {
    const next = options ? this.resolve(options) : this.clone();
    let base = Schema.prototype.describe.call(next, options)  as SchemaObjectDescription
    ...
  }

```

DaddyWarbucks:
This has been hard to test because things like Stackblitz, Codesandbox, etc struggle with `console.time` and `performance.now` kind of timers. So the only place that I could get a little test harness was directly in the browser and in node. But the following should work in Node and give a good example. Run this code in a file with access to Yup. You should notice some interesting results. The go into the node modules and make the changes made in this PR and you will see some substantial performance benefits.

I thought about making this a repo, but actually figured it would be easier for the end user if I just pasted it here.

```js
const yup = require('yup')

const person = {
  firstName: "Johnny",
  lastName: "Cash"
};

const people = Array(10000)
  .fill(0)
  .map((_, index) => {
    return {
      _id: index,
      createdAt: new Date(),
      updatedAt: new Date(),
      firstName: person.firstName + index,
      lastName: person.lastName + index,
      nickName: person.lastName + index,
      company: person.lastName + index,
      title: person.lastName + index,
      notes: person.lastName + index,
      emails: [
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false }
      ],
      phones: [
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false }
      ],
      dates: [
        { type: "HOME", value: new Date(), default: false },
        { type: "HOME", value: new Date(), default: false },
        { type: "HOME", value: new Date(), default: false }
      ],
      websites: [
        { type: "HOME", value: new Date(), default: false },
        { type: "HOME", value: new Date(), default: false },
        { type: "HOME", value: new Date(), default: false }
      ],
      socials: [
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false },
        { type: "HOME", value: "na@example.com", default: false }
      ]
    };
  });

console.log("Total People: ", people.length);

const syncSchema = yup.object({
  _id: yup.string().trim().matches("^[a-zA-Z0-9]*$").label("ID"),
  createdAt: yup.date().label("Created At"),
  updatedAt: yup.date().label("Updated At"),
  firstName: yup
    .string()
    .trim()
    .label("First Name"),
  lastName: yup.string().trim().label("Last Name"),
  nickName: yup.string().trim().label("Nickname"),
  company: yup.string().trim().label("Company"),
  title: yup.string().trim().label("Title"),
  avatar: yup.string().trim().label("Avatar"),
  emails: yup
    .array()
    .min(1)
    .label("Emails")
    .of(
      yup.object({
        type: yup.string().trim().required().label("Email Type"),
        value: yup.string().trim().email().required().label("Email"),
        default: yup.boolean().label("Default")
      })
    ),
  phones: yup
    .array()
    .min(1)
    .label("Phone Numbers")
    .of(
      yup.object({
        type: yup.string().trim().required().label("Phone Number Type"),
        value: yup.string().trim().required().label("Phone Number"),
        default: yup.boolean().label("Default")
      })
    ),
  dates: yup
    .array()
    .min(1)
    .label("Dates")
    .of(
      yup.object({
        type: yup.string().trim().required().label("Date Type"),
        value: yup.string().trim().required().label("Date"),
        default: yup.boolean().label("Default")
      })
    ),
  websites: yup
    .array()
    .min(1)
    .label("Websites")
    .of(
      yup.object({
        type: yup.string().trim().required().label("Website Type"),
        value: yup.string().trim().required().label("Website"),
        default: yup.boolean().label("Default")
      })
    ),
  socials: yup
    .array()
    .min(1)
    .label("Socials")
    .of(
      yup.object({
        type: yup.string().trim().required().label("Social Type"),
        value: yup
          .string()
          .transform((value, originalValue, ctx) => {
            if (ctx.isType(value) && value.startsWith("@")) {
              return value.substring(1);
            }
            return value;
          })
          .trim()
          .required()
          .label("Username"),
        default: yup.boolean().label("Default")
      })
    )
});

const asyncSchema = syncSchema.shape({
  firstName: yup
  .string()
  .trim()
  .label("First Name")
  .test({
    name: "async",
    exclusive: true,
    message: "Oh dear",
    test: async function () {
      return true;
    }
  })
})

const validateAsync = async (data) => {
  return asyncSchema.validate(data, { abortEarly: false, stripUnknown: true });
};

const validateSync = (data) => {
  return syncSchema.validateSync(data, { abortEarly: false, stripUnknown: true });
};

(async () => {
  console.time("validateAsync");
  await validateAsync(people[0]);
  console.timeEnd("validateAsync");

  console.time("validateSync");
  validateSync(people[0]);
  console.timeEnd("validateSync");

  console.time("validateSync.loop");
  for (let i = 0; i < people.length; i++) {
    const person = people[i];
    validateAsync(person);
  }
  console.timeEnd("validateSync.loop");

  console.time("validateAsync.loop");
  for (let i = 0; i < people.length; i++) {
    const person = people[i];
    await validateAsync(person);
  }
  console.timeEnd("validateAsync.loop");

  console.time("validateAsync.all");
  await Promise.allSettled(
    people.map(async (person, index) => {
      // console.time("multi" + index);
      return validateAsync(person);
      // console.timeEnd("multi" + index);
    })
  );
  console.timeEnd("validateAsync.all");
})();

```

src/array.ts
DaddyWarbucks: Another solution would be to remove this line: https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/schema.ts#L908
And then add a `describe` method to every type that does the resolve/clone itself before calling `super.describe()`. This is explicit and seems OK too.

Which doe you prefer @jquense?
- Use `describe` and `_describe` as the code is now
- Use `Schema.prototype.describe.call`
- Define and explicit `describe` method on each type that is responsible for resolving itself

jquense: what about:

```tsx
const next = this.resolve(options)
const base = this.withMutation(() => {
   return next.describe() // no options, shouldn't resolve
})

...
```

`withMutation` could be removed i am not sure why the root describe calls `clone()` maybe just to be safe with handing out mutable schema internals

DaddyWarbucks: I believe calling `next.describe()` would cause an infinite loop. I think I have a good solution, and I have updated the PR @jquense.
The code now resolves these types twice in the `describe` method. For example
```js
const next = (options ? this.resolve(options) : this).clone();
const base = super.describe(options) as SchemaInnerTypeDescription;
```
Then we can use `next.whatever` instead of `this.whatever` to use the new resolved schema to add additional information to the description. I'm not really concerned about any performance overhead there. I don't think the `describe` method is generally being used in performance critical areas, and the `clone` and `resolve` should be fast anyway.

jquense (approved):
yeah this is nice straightforward, works for me

lukefredrickson:
@DaddyWarbucks thanks for working on this — much appreciated!

lukefredrickson:
@jquense @DaddyWarbucks Sorry to be a bother, but is do you have an ETA for merging this work in?

DaddyWarbucks:
Thats above my pay grade.

PR #2072, opened 2023-07-29
#2043 performance improvement

tedeschia:
Resolves performance issues for large data validation described at #2043:

- Arrays of ValidationError
  - Mutate the array with splice instead of using concat
  - ValidationError class
- Add the new option "disableStackTrace" to stop capturing stack trace when instantiating ValidationError
  - Stop ValidationError extending Error, so no need to call super() when instantiating ValidationError

This PR also fix the issue on ValidationError that broke several tests in #2044


tedeschia:
Hi @jquense, do you think we could merge this PR? it solves all tests issues. Please let me know if I can do anything else to complete it. Thanks!

PR #2080, opened 2023-08-09
refactor(parse-iso-date): Convert to TypeScript

0livare:
The poor `parseIsoDate` util was the only file in the repo that hadn't been converted over to TypeScript back in 2020. In addition, it was written in a way that was hard to read and harder to understand. It also had the linter disabled for the entire file.

This pull request aims to fix all those issues, while maintaining exactly the same functionality. 

To ensure the same functionality, before making any changes I ported over the tests from the original source of the `parseIsoDate` function. Many of those tests no longer passed because of previous changes made to the function, but I modified them to match the module's current functionality. With the tests in place, I was then able to refactor the module and ensure that the functionality remained identical.


jquense (approved):
awesome thanks!

PR #2087, opened 2023-08-16
feat(string): Create .datetime()

0livare:
This PR builds on top of another currently open PR: #2080
(the diff here is larger than necessary because #2080 has not been merged)

`yup.date()` always attempts to parse input into a `Date` object. Sometimes that is undesirable and you just want the string to be left alone. Currently yup does not have a solution for that use case, but that's where `.datetime()` comes in!

With an API inspired by [zod.datetime](https://zod.dev/?id=iso-datetimes), `datetime` also provides greater control over the exact required format of the ISO string that `date` does.


0livare:
@jquense is this something you would consider?

README.md
```diff
@@ -1364,6 +1366,19 @@ Validates the value as a valid URL via a regex.
 
 Validates the value as a valid UUID via a regex.
 
+#### `string.datetime(options?: {message?: string | function, allowOffset?: boolean, precision?: number})`
+
+Validates the value as an ISO datetime via a regex. Defaults to UTC validation; timezone offsets are not permitted (see `options.allowOffset`).
+
+Unlike `.date()`, `datetime` will not convert the string to a `Date` object. `datetime` also provides greater customization over the required format of the datetime string than `date` does.
+
+`options.allowOffset`: Allow a time zone offset. False requires UTC 'Z' timezone. _(default: false)_
```
jquense, line 1375: why default to false for this?

src/string.ts
```diff
@@ -37,6 +45,14 @@ export type MatchOptions = {
   name?: string;
 };
 
+export type DateTimeOptions = {
+  message: Message<{ allowOffset?: boolean; precision?: number }>;
+  /** Allow a time zone offset. False requires UTC 'Z' timezone. (default: false) */
+  allowOffset?: boolean | null;
+  /** Require a certain sub-second precision on the date. (default: null -- any or no sub-second precision) */
+  precision?: number | null;
```
jquense, line 53: i'd have both of these options not be nullable

jquense:
yes open to this 👍

README.md
0livare: My thinking was that most of the time when you store an iso date in a DB (which in my mind will be the most common use case for this function) you would want them to be consistently formatted in UTC time for easy comparison.

jquense (approved):

jquense:
@0livare sorry this is taking so long but can you fix the build?

0livare:
@jquense build is fixed!  (no changes, I just rebased onto the latest master

jquense:
thanks!

PR #2091, opened 2023-08-21
export Reference class

### LINKED ISSUE #2083: Typing a retrieved Field Schema using yup.Reach()
I've finally decided to upgrade from  0.29 to 1.2 and faced some issues with it's types. I'm Looking for help to find a best solution for typing yup schemas.

I have components and functions, where I have an argument - a field schema (string/number/object, etc)

```ts
interface MyType {
  id: number;
  name: string;
}

const mySchema: ObjectSchema<MyType> = yup
  .object({
    id: yup.number().required(),
    name: yup.string().required(),
  })
  .required();

const idSchema = yup.reach(mySchema, "id");
const nameSchema = yup.reach(mySchema, "name");

function validateField(fieldSchema: Schema) {
  fieldSchema.validate("TEST");
}

validateField(idSchema); // <--- Here compiler complains, where in 0.29 it didn't
```

image

It seems that I got used to use `Schema` type everywhere, and now it doesn't satisfy neither Object schema value, nor any Field Schema.

lukasdruskis:
Closes #2083 

Exporting Reference class to be able to type retrieved Yup Schema.


jquense:
handled by: #2093, thanks!

Issue #2069, opened 2023-07-24
should all `nonNullable` receive a `msg?: Message` property?


csantos1113:
I would like to confirm if all `.nonNullable()` methods should receive a message or not as I see some methods allowing that to be configured but not others

### This methods allow `msg` prop in `nonNullable`

- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/schema.ts#L687
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/boolean.ts#L105-L106

### This methods do not allow `msg` prop in `nonNullable`

- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/array.ts#L309
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/date.ts#L138
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/mixed.ts#L76
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/number.ts#L195
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/object.ts#L115
- https://github.com/jquense/yup/blob/117843799e0fc3947f3583be7b60f309a25898d5/src/string.ts#L293

**To Reproduce**

https://codesandbox.io/s/yup1-0-2-non-nullable-4dy2rc?file=/src/index.ts

**Expected behavior**

Not really sure what is the expected behavior. I'd have expected that all types can be combined with `nonNullable()`, and that if I use `.nonNullable()` I can configure the message.

**Platform (please complete the following information):**
 - Browser: ALL
 - Version: yup 1.0.2

sadik-malik:
I was trying to use yup.object().nonNullable() but can't pass the customized message. 

For i18n purpose, allowing developers to pass custom message in nonNullable would be helpful.

PR #2119, opened 2023-10-07
fix: add optional message to nonNullable schema methods

ranquild:
Fixes https://github.com/jquense/yup/issues/2069

Adds an optional `msg?: Message` argument to `nonNullable` method on every schema object. The argument is already supported by the [base Schema class](https://github.com/jquense/yup/blob/master/src/schema.ts#L713), and we only need to pass it properly with TypeScript.

Also updates `nullable` and `nonNullable` signatures in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,8 +111,8 @@ const parsedUser = await userSchema.validate(
     - [`Schema.withMutation(builder: (current: Schema) => void): void`](#schemawithmutationbuilder-current-schema--void-void)
     - [`Schema.default(value: any): Schema`](#schemadefaultvalue-any-schema)
     - [`Schema.getDefault(options?: object): Any`](#schemagetdefaultoptions-object-any)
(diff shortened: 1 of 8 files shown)


ranquild:
@jquense could you take a look, please?

lmarvaud:
Hello,

It seem that the typing of "notNullable" and "nullable" had been swaped. In `Schema.ts`, inherited by few validators, only nonNullable takes a parameter.
 https://github.com/jquense/yup/blob/5db2b2fdd6276ff8ae8c15e8aa9d3aa9710ad4ac/src/schema.ts#L710-L715
 
Actually wouldn't it be cleaner to add those types in the [Schema interface](https://github.com/jquense/yup/blob/5db2b2fdd6276ff8ae8c15e8aa9d3aa9710ad4ac/src/schema.ts#L956), to avoid code duplication and those type of error ?

jquense (approved):

jquense:
thanks!

PR #2143, opened 2023-12-06
renamed miss spelled method of numbers

sarathkumarsasi:
Renamed miss spelled method for numbers


jquense:
Hey there thanks, but this isn't misspelled, it intentionally shares a name with the the other `max` method, as they are different ways of saying the same thing, so only one should apply at a time

Issue #2156, opened 2024-01-03
validateSync() returns invalid paths for nested errors in some cases where field names contain "." (period)


ir0nCr055:
**Describe the bug**

The `validateSync()` function may return invalid error paths for some nested object field names containing "." (period).

Namely, the invalid paths returned are missing quotes.

**To Reproduce**

Given the following schema and object...

```
const schema = yup.object({
    "i.t._admin_name": yup.string().required("Enter a valid name"),
    "i.t._admin_email": yup
      .string()
      .required("Enter a valid email")
      .test("valid_email", "Email is invalid", (val) => {
        return false; // This test always fails to simplify the demonstration
      }),
  });

  // This example demonstrates the bug - we expect errors for both keys.
  const obj = {
    "i.t._admin_name": "",
    "i.t._admin_email": "invalid email",
  };
```

Evaluating `schema.validateSync(obj, { abortEarly: false })` throws a validationError containing the following paths:

  - `["i.t._admin_name"]`
  - `[i.t._admin_email]`

This test case is demonstrated here: https://codesandbox.io/p/devbox/yh65qd

**Expected behavior**

We'd expect the paths to be:

  - `["i.t._admin_name"]`
  - `["i.t._admin_email"]`

**Platform**
 
All tested browsers, versions including Chrome, Firefox.

**Additional context**

Bug discovered while integrating Yup with Formik and React.

PR #2157, opened 2024-01-03
Fixes a typo in the asNestedTest() path rendering logic causing issue #2156

ir0nCr055:
Fixes a typo in `asNestedTest()` which is causing #2156


jquense (approved):
thanks!

PR #2164, opened 2024-01-15
Moved the localized object into it's own file

trajano:
This will provide a example for i18n to translate as it is a single object that needs to be manipulated and separated from the type definition.

In addition the `as` was removed as it type coercion rather than a type check


jquense:
Thanks for the PR! I appreciate the effort but I am going to pass on this split. I don't actually want folks to contribute translations as i wouldn't be able to keep them updated or up to date, and for users who need to translate it's easy enough to just copy the part of the file needed. thanks again

PR #2203, opened 2024-04-19
Fixed logic in "42" tests

rmc29:
The logic in the "42" tests was backwards. It was checking that the number didn't equal 42, when it should check the number does equal 42.

### DIFF
--- a/README.md
+++ b/README.md
@@ -779,7 +779,7 @@ For instance this will work:
 let schema = number().test(
   'is-42',
   "this isn't the number i want",
-  (value) => value != 42,
+  (value) => value === 42,
 );
 
 schema.validateSync(23); // throws ValidationError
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -779,7 +779,7 @@ For instance this will work:
 let schema = number().test(
   'is-42',
   "this isn't the number i want",
-  (value) => value != 42,
+  (value) => value === 42,
```
jquense, line 782:
```suggestion
  (value) => value == null || value === 42,
```

README.md
```diff
@@ -789,7 +789,7 @@ however this will not:
 
 ```js
 let schema = number().test('is-42', "this isn't the number i want", (value) =>
-  Promise.resolve(value != 42),
+  Promise.resolve(value === 42),
```
jquense, line 792:
```suggestion
  Promise.resolve(value == null || value === 42,),
```

README.md
rmc29: Are you saying you want the test to pass if the value is null?

jquense: if its undefined or null. that is handled by `required`, `defined` etc

PR #2217, opened 2024-05-14
fix(regex): update the email regex for more cases

cookieMonsterDev:
The yup.string().email() works via regex but is does not cover cases like:

- test-user@abccom
- test-user@abc.com
- test.user.@abc.com
- test..user@abc.com
- test.user@abc–88.com
- test.user@abc.r
- 0@g.q

This PR resolves this issue.


jquense:
Thanks for the PR but the current behavior is intentionally matching browser spec for email validation. You can read more about the reasons and justifications in the b1 issue and other issues

PR #2227, opened 2024-07-01
let/const consistency in README: let preference

dearlordylord:
I noticed let and const are used interchangeably in README. For consistency, I'd like to propose to use either/or. The previous PR https://github.com/jquense/yup/pull/2226 was about changing all `let`s to `const`s

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ let userSchema = object({
 });
 
 // parse and assert validity
-const user = await userSchema.validate(await fetchUser());
+let user = await userSchema.validate(await fetchUser());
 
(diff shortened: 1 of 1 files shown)


alexandermirzoyan:
Why to use `let` if in the code snippets variables are not being changed?

dearlordylord:
That's a big topic (why not use DeepReadonly<> for every structure then?), but the goal of the PR is consistency. There's a satellite PR that makes everything `const`. The point is that in the README, `let` and `const` are intermixed for no apparent reason.

jquense (approved):

jquense:
thanks!

PR #2250, opened 2024-09-30
Fix `ValidationError.formatError()` clobbering `path` param

shaungrady:
## Changes
- `ValidationError.formatError()` overwrites `path` with a more friendly `label` string (if defined) for more friendly error message interpolation; however, this causes the `path` data to be lost entirely.
  - Rather than undertake refactoring this logic and cause a breaking change, where using alternative values for interpolating `${path}` within error strings should arguably be a concern that's left to the `message` function, I opted to simply create a new `params` property—`originalPath`—for recording this value.
- Update `MessageParams` type with `originalPath: string;` property
- Add `MessageParams` and `AnyMessageParams` to library exports
  - I couldn't see a reason why these types are private, and I'd certainly love to make use of them.

### DIFF
--- a/src/ValidationError.ts
+++ b/src/ValidationError.ts
@@ -63,8 +63,11 @@ export default class ValidationError extends Error {
     message: string | ((params: Params) => string) | unknown,
     params: Params,
   ) {
+    // Attempt to make the path more friendly for error message interpolation.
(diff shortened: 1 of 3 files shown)


jquense (approved):

jquense:
thanks!

PR #2258, opened 2024-12-04
feat: Implement standard schema interface

### LINKED ISSUE #2255: Support Standard-schema API?
Hello,

[standard-schema](https://github.com/standard-schema/standard-schema) is a community effort to build a standard interface for schema libraries without any runtime dependencies, just by implementing a type.

I think the community can benefit from having yup implement the standard schema interface which will improve compatibility and easy migration from and to participating schema providers.

Being a maintainer of a popular form library and with another project on the way, it would be much easier to bake in type inference and validation capabilities without requiring users to install a resolver.

I think Yup already has all the parts needed, it is just a matter of conforming to the standard schema API.

I realize you may not have time to add this in. I can PR this, but I would like to get your thoughts before I dig in.

logaretm:
# What

This is a PR that implements the [standard schema interface](https://github.com/standard-schema/standard-schema) for yup.

**Motivation:** Standard schema is an initiative that is aimed to get all Typed schema providers to support a unified interface that will enable various tools and 3rd party libraries to seamlessly integrate with "a schema provider" which reduces friction when migrating from one provider to the other, and allows mixing these providers if one is suited for a task more than the rest. Another benefit is it allows 3rd party authors to ditch the "resolver" pattern which in turn reduces their users' friction.

We discussed this briefly in #2255 with no clear green light given on this, so I understand if this is something you don't want the burden of maintaining. I already did the work locally to play with it, since I was planning to build a yup standard schema adapter anyways. So, I thought I might as well just PR it.

I made sure to add tests and type tests as well, happy to address any issues in this PR.

# How

To avoid adding `@standard-schema/spec` as a dependency, I copied the types which is the [recommended approach](https://github.com/standard-schema/standard-schema#schema-library). I then make sure it is aligned with the spec by testing it against the spec which I added as a dev dependency.

I also had to implement a path splitter to make the paths compatible with the standard schema output.

#### Bundle size diff (~3%)

| Label | Size Before | Size After |
| --- | --- | --- |
| Bundle Size | 72.6 KB |  75.73 KB |
| Minified Size | 39.07 KB | 40.08 KB |
| Gzipped Size | 11.44 KB | 11.79 KB |

closes #2255

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -34,6 +34,11 @@ import isAbsent from './util/isAbsent';
 import type { Flags, Maybe, ResolveFlags, _ } from './util/types';
 import toArray from './util/toArray';
 import cloneDeep from './util/cloneDeep';
+import {
+  createStandardSchemaProps,
+  type StandardSchema,
+  type StandardSchemaProps,
+} from './standardSchema';
 
 export type SchemaSpec<TDefault> = {
   coerce: boolean;
@@ -148,7 +153,9 @@ export default abstract class Schema<
   TContext = any,
   TDefault = any,
   TFlags extends Flags = '',
-> implements ISchema<TType, TContext, TFlags, TDefault>
+> implements
+    ISchema<TType, TContext, TFlags, TDefault>,
+    StandardSchema<TType, ResolveFlags<TType, TFlags, TDefault>>
 {
   readonly type: string;
 
@@ -174,6 +181,11 @@ export default abstract class Schema<
 
   protected exclusiveTests: Record<string, boolean> = Object.create(null);
   protected _typeCheck: (value: any) => value is NonNullable<TType>;
+
+  public '~standard': StandardSchemaProps<
+    TType,
+    ResolveFlags<TType, TFlags, TDefault>
+  >;
 
   spec: SchemaSpec<any>;
 
@@ -203,6 +215,11 @@ export default abstract class Schema<
     this.withMutation((s) => {
       s.nonNullable();
     });
+
+    this['~standard'] = createStandardSchemaProps<
+      TType,
+      ResolveFlags<TType, TFlags, TDefault>
+    >(this);
   }
 
   // TODO: remove
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -35,7 +35,8 @@ import type { Flags, Maybe, ResolveFlags, _ } from './util/types';
 import toArray from './util/toArray';
 import cloneDeep from './util/cloneDeep';
 import {
-  createStandardSchemaProps,
+  issuesFromValidationError,
+  StandardResult,
   type StandardSchema,
   type StandardSchemaProps,
 } from './standardSchema';
@@ -182,11 +183,6 @@ export default abstract class Schema<
   protected exclusiveTests: Record<string, boolean> = Object.create(null);
   protected _typeCheck: (value: any) => value is NonNullable<TType>;
 
-  public '~standard': StandardSchemaProps<
-    TType,
-    ResolveFlags<TType, TFlags, TDefault>
-  >;
-
   spec: SchemaSpec<any>;
 
   constructor(options: SchemaOptions<TType, any>) {
@@ -215,11 +211,6 @@ export default abstract class Schema<
     this.withMutation((s) => {
       s.nonNullable();
     });
-
-    this['~standard'] = createStandardSchemaProps<
-      TType,
-      ResolveFlags<TType, TFlags, TDefault>
-    >(this);
   }
 
   // TODO: remove
@@ -979,6 +970,41 @@ export default abstract class Schema<
     };
 
     return description;
+  }
+
+  get ['~standard']() {
+    const schema = this;
+
+    const standard: StandardSchemaProps<
+      TType,
+      ResolveFlags<TType, TFlags, TDefault>
+    > = {
+      version: 1,
+      vendor: 'yup',
+      async validate(
+        value: unknown,
+      ): Promise<StandardResult<ResolveFlags<TType, TFlags, TDefault>>> {
+        try {
+          const result = await schema.validate(value, {
+            abortEarly: false,
+          });
+
+          return {
+            value: result as ResolveFlags<TType, TFlags, TDefault>,
+          };
+        } catch (err) {
+          if (err instanceof ValidationError) {
+            return {
+              issues: issuesFromValidationError(err),
+            };
+          }
+
+          throw err;
+        }
+      },
+    };
+
+    return standard;
   }
 }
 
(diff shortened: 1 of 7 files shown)


jquense:
I see this and appreciate your work here, i'm trying to get some time to dedicate to looking at it!

test/standardSchema.ts
```diff
@@ -0,0 +1,67 @@
...
+
+test('is compatible with standard schema', () => {
+  expect(verifyStandardSchema(string())).toBe(true);
+  expect(verifyStandardSchema(number())).toBe(true);
+  expect(verifyStandardSchema(array())).toBe(true);
+  expect(verifyStandardSchema(bool())).toBe(true);
+  expect(verifyStandardSchema(object())).toBe(true);
+  expect(verifyStandardSchema(date())).toBe(true);
+  expect(verifyStandardSchema(mixed())).toBe(true);
+  expect(verifyStandardSchema(tuple([mixed()]))).toBe(true);
```
jquense, line 31: this is missing `lazy`

src/schema.ts
```diff
@@ -202,6 +214,11 @@ export default abstract class Schema<
     this.withMutation((s) => {
       s.nonNullable();
     });
+
+    this['~standard'] = createStandardSchemaProps<
```
jquense, line 218: this won't work correctly, b/c it closes over the `schema` instance. Most method calls in yup create new schema instances via cloning and the `validate` method won't keep up with them, e.g. `string().required()['~standard'].validate` will be the nonRequired schema instance

src/standardSchema.ts
```diff
@@ -0,0 +1,176 @@
...
+  // Array to store the final path segments
+  const segments: string[] = [];
+  // Buffer for building the current segment
+  let currentSegment = '';
+  // Track if we're inside square brackets (array/property access)
+  let inBrackets = false;
+  // Track if we're inside quotes (for property names with special chars)
+  let inQuotes = false;
+
+  for (let i = 0; i < path.length; i++) {
```
jquense, line 96: what is going on with path logic? There are already utilities for splitting and traversing yup's path's in utils, see the `reach` utility

src/standardSchema.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    if (char === ']' && !inQuotes) {
+      if (currentSegment) {
+        // Handle numeric indices (e.g. arr[0])
+        if (/^\d+$/.test(currentSegment)) {
+          segments.push(currentSegment);
+        } else {
+          // Handle quoted property names (e.g. obj["foo.bar"])
+          segments.push(currentSegment.replace(/^"|"$/g, ''));
```
jquense, line 116: i don't think yup even handles these cases, i'm not sure it makes sense to do it here, it's misleading, b/c validate will still likely treat these as actual path segments

test/standardSchema.ts
jquense, line 1: can we add lot more diverse and exhaustive tests around this interface?

src/schema.ts
logaretm: You are right, I re-implemented them as getters now which should carry over well with cloning. I added a couple of tests to test that as well.

test/standardSchema.ts
logaretm: Added it, along with a special test for it.

test/standardSchema.ts
logaretm: Absolutely, I'm relying on the other parts working as expected because this only adapts the interface to yup's current implementation so I was only testing that layer.

Happy to add more tests, what kind of tests would you like to see? General ones for each schema?

src/standardSchema.ts
logaretm: Here I'm not traversing the schema path which is what I think `reach` and `getIn` are for?

What I'm doing here is I'm converting error path from as single string value to an array of strings/symbols which is what the standard schema interface requires. 

For example these are the output paths from ValidationErrors and next to them is how they should look like in standard schema:

| Error Message Path | Array Format |
|-------------------|--------------|
| obj.foo | ['obj', 'foo'] |
| obj["not.obj.nested"] | ['obj', 'not.obj.nested'] |
| arr[0].foo | ['arr', '0', 'foo'] |
| arr[0]["not.array.nested"] | ['arr', '0', 'not.array.nested'] |
| ["not.a.field"] | ['not.a.field'] |

I scanned the codebase quickly, do we have utilities for this kind of thing?

src/standardSchema.ts
logaretm: Right, I didn't really encounter this in all my time using `yup` so I didn't consider if it was supported or not. What do you think I should do then?

logaretm:
Thanks for taking a look, I addressed some of the issues and added clarification to your comments on the path stuff. Happy to keep iterating with your guidance.

anton-patrushev:
What's the plan here? Is there a chance this could be merged in the upcoming future?
cc: @logaretm 

And btw, thanks for looking into this @logaretm 🙌 ❤️ 

I would love to see this PR merged to get myself unblocked with the upgrade TanStack Form and still have my yup validator integrations working and postponing, for now, the change and migration to a different object validation solutions like zod, etc.

test/standardSchema.ts
jquense: I would take a number of the other object, array, and lazy tests and test that they produce the same results as the primary yup interface, at least in turns of valid vs invalid.

The ones i'm thinking about in particular are around `concat`, `extend`, as well as conditionals like `when` and and references

logaretm:
@franciscohermida I halted the release that would drop it to find time for this one.

@jquense No worries, I'm too familiar with the open-source life 😂 I will pick this up with your guidance and I will try to address the issues you raised.

test/standardSchema.ts
logaretm: Just got back around to doing this, so I had AI add test cases for various random examples for arrays, objects, and lazy to ensure the standard schema interface produces the same errors as the main API.

I also had it write tests for conditionals with `when` and `extend` and `concat` APIs and the implementation seems to hold in these cases as well.

Let me know if you think this is acceptable and thank you for following up!

logaretm:
@jquense I added more tests around the areas you requested, I used AI to do the bulk of the work. Hopefully this should be acceptable, the implementation seems to be holding well.

jquense:
Looks great, i need to figure out why CI pipeline is broken. 

Can you add a few docs to the readme just noting that it's standard schema compatible

logaretm:
Great, thank you! I added a note and a section to the README 🙌

jquense:
released in `v1.70` Thank you very much @logaretm 🎉

logaretm:
@dilanluna I don't think it was a requirement for the standard schema spec, at least I'm not aware of it.

I believe we could parse them as numbers relatively simple since we know when it is an array indices:

https://github.com/jquense/yup/blob/cdb4bffb8912d2b47fa0dbe988ee2c8af9736609/src/standardSchema.ts#L77-L81

PR #2277, opened 2025-03-12
feat: update array validation messages for pluralization

sbourouis:
The default validation messages always showed "items" as plural for arrays validation even though the number was 0 or 1.
I've updated them to show "item" when the value is "1" and "items" otherwise.

"this field must have less than or equal to 1 items" becomes "this field must have less than or equal to 1 item"
"this field must have at least 1 items" becomes "this field must have at least 1 item"
"this must have 1 items" becomes "this must have 1 item"

### DIFF
--- a/src/locale.ts
+++ b/src/locale.ts
@@ -139,9 +139,12 @@ export let object: Required<ObjectLocale> = {
 };
 
 export let array: Required<ArrayLocale> = {
-  min: '${path} field must have at least ${min} items',
-  max: '${path} field must have less than or equal to ${max} items',
-  length: '${path} must have ${length} items',
+  min: ({ path, min }) =>
+    `${path} field must have at least ${min} item${!min || min === 1 ? '' : 's'}`,
+  max: ({ path, max }) =>
+    `${path} field must have less than or equal to ${max} item${!max || max === 1 ? '' : 's'}`,
+  length: ({ path, length }) =>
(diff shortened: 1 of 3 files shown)


src/locale.ts
```diff
@@ -139,9 +139,12 @@ export let object: Required<ObjectLocale> = {
...
 export let array: Required<ArrayLocale> = {
-  min: '${path} field must have at least ${min} items',
-  max: '${path} field must have less than or equal to ${max} items',
-  length: '${path} must have ${length} items',
+  min: ({ path, min }) =>
+    `${path} field must have at least ${min} item${!min || min === 1 ? '' : 's'}`,
+  max: ({ path, max }) =>
+    `${path} field must have less than or equal to ${max} item${!max || max === 1 ? '' : 's'}`,
+  length: ({ path, length }) =>
+    `${path} must have ${length} item${!length || length === 1 ? '' : 's'}`,
```
sbourouis, line 147:
```suggestion
    `${path} field must have at least ${min} item${min === 1 ? '' : 's'}`,
  max: ({ path, max }) =>
    `${path} field must have less than or equal to ${max} item${max === 1 ? '' : 's'}`,
  length: ({ path, length }) =>
    `${path} must have ${length} item${length === 1 ? '' : 's'}`,
```

jquense:
This is a nice change however, it's unfortunately a breaking one, since error messages are often depended on in code. I don't want to break someone's app due to something like this. next major 🙏 though

PR #2281, opened 2025-03-27
Fix: Ensure validation errors are sorted by schema insertion order

### LINKED ISSUE #2279: Schema objects aren't ordered correctly in the error list
Schema objects aren't ordered correctly in the error list. Objects should appear in the list in the same order they are presented in the schema. This works fine with fields, but if you have multiple objects on the same level the ordering of them goes wrong.

Below is a snippet of code you can test this with. It will log the errors out.

```
import * as yup from 'yup' 

const schema = yup.object({
  testObjects: yup.object().shape({
    object: yup.object().shape({
      field: yup.string().required(),
    }),
    objectC: yup.object().shape({
      field: yup.string().required(),
    }),
    objectB: yup.object().shape({
      field: yup.string().required(),
    }),
  }),
})

// Example data to validate
const data = {
  testObjects: {
    object: {
      field: '',
    },
    objectB: {
      field: '',
    },
    objectC: {
      field: '',
    },
  },
}

async function validateData() {
  try {
    await schema.validate(data, { abortEarly: false }) // `abortEarly: false` to get all errors
    console.log('Validation successful')
  } catch (error) {
    if (error instanceof yup.ValidationError) {
      const errors = error.inner.map((e) => ({
        path: e.path,
        message: e.message,
      }))
      console.log('Validation Errors:', errors)
    }
  }
}

validateData()
```

The above code will log the following and as you can see `testObjects.object` is last on the list even though its first in the schema. Its not ordering them alphabetically either. Changing `testObjects.objectB` and `testObjects.objectC` order will have effect though, but still `testObjects.object` will remain last.

```
[
    {
        "path": "testObjects.objectB.field",
        "message": "testObjects.objectB.field is a required field"
    },
    {
        "path": "testObjects.objectC.field",
        "message": "testObjects.objectC.field is a required field"
    },
    {
        "path": "testObjects.object.field",
        "message": "testObjects.object.field is a required field"
    }
]
```

Expected behaviour is that all errors are listed in the order they are in the schema.

This is an issue because you can't rely on the top down ordering of the fields anymore when for example pointing out errors on a form.

**Platform:**
 - Chrome
 - Version 134.0.6998.89

severikupari1:
### Summary

This PR addresses #2279 by ensuring that validation errors are returned in the same order as fields are defined in the schema — particularly when multiple nested objects exist at the same level. Previously, validation errors could appear in reverse or unpredictable order due to internal traversal logic.

### Problem

When validating schemas with `abortEarly: false`, Yup collects errors from nested objects but does not guarantee that those errors respect the declaration order within the schema. For example:

```ts
const schema = yup.object({
  testObjects: yup.object().shape({
    object: yup.object({ field: yup.string().required() }),
    objectC: yup.object({ field: yup.string().required() }),
    objectB: yup.object({ field: yup.string().required() }),
  }),
})
```

Produces this incorrect error order:
```json
[
  { "path": "testObjects.objectB.field", "message": "..." },
  { "path": "testObjects.objectC.field", "message": "..." },
  { "path": "testObjects.object.field", "message": "..." }
]
```

Fix

- This PR implements the following:
- Recursively collects schema field paths (dot notation) in declaration order.
- Adds inline logic in runTests() to dynamically sort nestedErrors using a comparator built from those field paths.
- Introduces a depth limit (default 25) to prevent excessive or infinite recursion.
- Uses a WeakMap to cache paths per schema for performance.

Highlights
- No changes to schema APIs (_sortErrors or options) — logic is entirely contained within runTests().

No breaking changes.

Inline and scoped: only affects the error collection phase when abortEarly: false.

### Example After Fix
```json
[
  { "path": "testObjects.object.field", "message": "..." },
  { "path": "testObjects.objectB.field", "message": "..." },
  { "path": "testObjects.objectC.field", "message": "..." }
]
```
### Closes
#2279

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -33,6 +33,7 @@ import isAbsent from './util/isAbsent';
 import type { Flags, Maybe, ResolveFlags, _ } from './util/types';
 import toArray from './util/toArray';
 import cloneDeep from './util/cloneDeep';
+import sortByKeyOrder, { collectFieldPaths } from './util/sortByKeyOrder';
 
 export type SchemaSpec<TDefault> = {
   coerce: boolean;
@@ -489,13 +490,13 @@ export default abstract class Schema<
     let nestedErrors = [] as ValidationError[];
 
     if (!count) return nextOnce([]);
-
+    const self = this as any; // Capture `this`
     let args = {
       value,
       originalValue,
       path,
       options,
-      schema: this,
+      schema: self,
     };
 
     for (let i = 0; i < tests.length; i++) {
@@ -508,6 +509,9 @@ export default abstract class Schema<
             : nestedErrors.push(err);
         }
         if (--count <= 0) {
+          // Create sort order from schema
+          const sortOrder = collectFieldPaths(self, path);
+          nestedErrors.sort(sortByKeyOrder(sortOrder));
           nextOnce(nestedErrors);
         }
       });
--- a/src/util/sortByKeyOrder.ts
+++ b/src/util/sortByKeyOrder.ts
@@ -16,3 +16,85 @@ export default function sortByKeyOrder(keys: readonly string[]) {
     return findIndex(keys, a) - findIndex(keys, b);
   };
 }
+
+// WeakMap used for caching schema path results per schema instance.
+// This ensures no redundant recursion and avoids memory leaks,
+// since WeakMap entries are garbage collected when schema objects are no longer referenced.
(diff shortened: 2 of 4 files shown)


jquense:
hey there, thanks for attempting to fix this. The solution here is too complex for a simple sort order convenience. The bug is just in the `findIndex` function in the existing code. We just need to fix it so it's smarter than looking at specific substrings

kiikka:
Has this issue been fixed?

PR #2296, opened 2025-07-19
fix: typefix the string.ensure() method

vmanera:
The `string.ensure()` method applies`.default('')`, however this information does not flow to typescript. This happens because the explicit return type of the method causes a loss of information about the transforms that happen within. The typing changes I have in this PR will ensure the type information flows through correctly.

Pragmatically, this issues effects the use of `Schema.getDefault()` method. Without this fix, any strings using `.ensure()` will have a type of `undefined`, but a runtime value of `""`.

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -265,10 +265,10 @@ export default class StringSchema<
   }
 
   //-- transforms --
-  ensure(): StringSchema<NonNullable<TType>> {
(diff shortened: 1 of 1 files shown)


deansimcox (approved):
Love your work.
Happy to see the `as any` on line 271 has been removed.
5 ⭐ 's

PR #2305, opened 2025-08-20
Pass proper options to array casting.

### LINKED ISSUE #2187: Parent selector (`TestContext.from`) skips array schemas.
The `TestContext.from` property is not documented yet. It was introduced in #556 without any documentation or typescript support.  In an attempt to document this I stumbled upon an irregularity.

-> `from` seems to skip parents that are arrays. 

This behavior looks like it is not accidental though since it is used in the tests: 

https://github.com/jquense/yup/blob/5c77e0d4f9373151bcf0cd558c95986b6e4800d7/test/mixed.ts#L487-L524

## Actual
```js
expect(finalFrom[0].value).toEqual(testFixture.second[finalOptions.index]);
expect(finalFrom[0].schema).toBe(third);
expect(finalFrom[1].value).toBe(testFixture);
expect(finalFrom[1].schema).toBe(first);
```

## Expected
Reading the information from the PR I expected following test cases to be valid.
```js
expect(finalFrom[0].value).toEqual(testFixture.second[finalOptions.index]);
expect(finalFrom[0].schema).toBe(third);
expect(finalFrom[1].value).toBe(testFixture.second);
expect(finalFrom[1].schema).toBe(second);
expect(finalFrom[2].value).toBe(testFixture);
expect(finalFrom[2].schema).toBe(first);
```

How could we describe current behavior?
Is the current behavior what we expect the `from` property to behave like?

### LINKED ISSUE #2198: .when() context is inconsistent
**Describe the bug**
When using the third argument of `.when()` (aka `testContext`), the transform does not run as expected. Specifically I am trying to use the `testContext.index` of an `array` schema to access items retrieved from the `when`.

**To Reproduce**
I cannot, to save my life, figure out how to open the test panel on Codesandbox. I also was pulling my hair out because Codesandbox was running the code twice and this error is so weird in general. Instead I used StackBlitz. To access the console, inspect the page and it will bring up a separate Chrome DevTools (not embedded like Codesandbox).

https://stackblitz.com/edit/vitejs-vite-r7lfvf?file=main.js

Switch between commenting lines `11` and `14`. Although both of these lines of code should do the exact same thing, line `11` does not work.

**Expected behavior**
I can use the `testContext.index` to retrieve things from context.

 - Browser: Chrome
 - Version: 1.4

**Additional context**
I am happy to work on this and can open a PR. But I am unclear on where to even start.

DaddyWarbucks:
This PR aims to consolidate the options passed to arrays and tuples when casting/validating.

When validating tuples/arrays, yup passes `index`, `value`, `originalValue`, and `parent` to nested tests via `asNestedTest`. But when casting, these properties are either not passed down or passed down incorrectly. For example, `parent` currently is incorrect for nested casting. The `parent` should be the array that this item is inside.

This is particularly helpful when using `.when()` on inner types.

```js
const itemSchema = string().when([], (v, schema, textContext) => {
  // do something with the textContext.index
})

const arraySchema = array(itemSchema)

await arraySchema.validate(['Johnny', 'Cash'])
```

For objects, there is no functional change to the code. I removed a `TODO` comment and reused a variable for clarity. The `TODO` stated that casting the underlying field should use `_cast` because the field schema has already been resolved and using `cast` would cause double resolving. But I don't believe that is the case. The `field.resolve` returns a new schema with no `conditions`. So when calling `field.cast`, it does call `this.resolve` again but there are no conditions to resolve so its a noop.

Closes #2198 
May also close #2187 ?

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -83,6 +83,10 @@ export interface CastOptions<C = {}> {
   stripUnknown?: boolean;
   // XXX: should be private?
   path?: string;
+  index?: number;
+  key?: string;
+  originalValue?: any;
+  value?: any;
   resolved?: boolean;
 }
 
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -378,8 +378,8 @@ export default abstract class Schema<
     options: CastOptions<TContext> | CastOptionalityOptions<TContext> = {},
   ): this['__outputType'] {
     let resolvedSchema = this.resolve({
-      value,
       ...options,
+      value,
       // parent: options.parent,
       // context: options.context,
     });
--- a/src/array.ts
+++ b/src/array.ts
@@ -81,11 +81,11 @@ export default class ArraySchema<
     const castArray = value.map((v, idx) => {
       const castElement = this.innerType!.cast(v, {
         ..._opts,
+        path: `${_opts.path || ''}[${idx}]`,
         parent: value,
         originalValue: v,
         value: v,
         index: idx,
-        path: `${_opts.path || ''}[${idx}]`,
 
       });
       
--- a/src/array.ts
+++ b/src/array.ts
@@ -81,8 +81,14 @@ export default class ArraySchema<
     const castArray = value.map((v, idx) => {
       const castElement = this.innerType!.cast(v, {
         ..._opts,
+        parent: value,
+        originalValue: v,
+        value: v,
+        index: idx,
         path: `${_opts.path || ''}[${idx}]`,
+
       });
+      
       if (castElement !== v) {
         isChanged = true;
       }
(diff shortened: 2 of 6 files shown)


DaddyWarbucks:
This is passing tests locally. I am not sure whats up with the build pipeline for testing here.

jquense:
thanks!

src/schema.ts
```diff
@@ -83,6 +83,10 @@ export interface CastOptions<C = {}> {
   stripUnknown?: boolean;
   // XXX: should be private?
   path?: string;
+  index?: number;
+  key?: string;
+  originalValue?: any;
+  value?: any;
```
jquense, line 89: sorry taking a closer look, why are we passing value and originalValue here. these are conflated with other values that are used in cast

DaddyWarbucks: I believe they were passed to be more consistent with the`testOptions` that are passed to `resolve` and `_validate` in `asNestedTest`. 
See: https://github.com/jquense/yup/blob/ff31eee8a2b10c938bb3a22544d9e33957a8cf01/src/schema.ts#L546

Issue #2303, opened 2025-08-04
`trim`, `lowercase`, `uppercase`: tries to do this even if there is no string and gets a TypeError


fir4tozden:
I previously opened an issue for this, but it was ignored. If `trim()`, `lowercase()`, or `uppercase()` is used in a `string()`, a TypeError error is thrown during validation if the incoming data is not a string.

### Everything here is as it should be ✅

<img width="1280" height="560" alt="Image" src="https://github.com/user-attachments/assets/eec50fd1-0c40-4f29-907c-b2f2d9eeeab8" />

### Everything here is as it should be ✅

<img width="1280" height="560" alt="Image" src="https://github.com/user-attachments/assets/97e1bc07-1e27-4670-8035-2bea4534ed62" />

### But it wasn't supposed to be like this ❌

<img width="1280" height="698" alt="Image" src="https://github.com/user-attachments/assets/3333a75d-79ed-4cf9-b68d-45e20b781817" />

### Temporary solution 🔧

We can use the `transform()` for now. However, this doesn't mean I should ignore the problem above.

```javascript
import * as Yup from "yup";

const schema = Yup.string()
  .transform(property => typeof property === "string" ? property.trim() : property)
  .min(3);

console.log(schema.validateSync({}));
```

mahmoodhamdi:
This happens because transform functions like trim(), lowercase(), and uppercase() run before type validation. When the value is not a string (e.g. undefined or a number), calling .trim() on it throws a TypeError.

As a workaround, you can add a typeError check before the transforms:

```ts
yup.string().typeError('Must be a string').trim().required()
```

Or use transform() with a type guard:

```ts
yup.string().transform((value) => {
  return typeof value === 'string' ? value.trim() : value;
}).required()
```

The underlying issue is that yup runs transforms in the pipeline before validating the type, which seems like a bug. The type check should run first, and transforms should only apply to values that match the expected type.

PR #2310, opened 2025-10-01
fix: checking whether the value checked for `.trim()`, `.lowercase()` and `.uppercase()` is a string

fir4tozden:
The `.trim()`, `.lowercase()`, and `.uppercase()` methods do not check whether a value is a string. Therefore, the application crashes if the entered value is not a string.

I previously opened an [issue](https://github.com/jquense/yup/issues/2303) for this, but it didn't receive much attention. This is a critical issue, and I wanted to resolve it. 

### Everything here is as it should be ✅

473884983-eec50fd1-0c40-4f29-907c-b2f2d9eeeab8

### Everything here is as it should be ✅

473885038-97e1bc07-1e27-4670-8035-2bea4534ed62

### But it wasn't supposed to be like this ❌

473885162-3333a75d-79ed-4cf9-b68d-45e20b781817

### DIFF
--- a/src/string.ts
+++ b/src/string.ts
@@ -272,7 +272,7 @@ export default class StringSchema<
   }
 
   trim(message = locale.trim) {
-    return this.transform((val) => (val != null ? val.trim() : val)).test({
+    return this.transform((val) => (typeof val === 'string' ? val.trim() : val)).test({
       message,
       name: 'trim',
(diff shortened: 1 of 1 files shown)


fir4tozden:
Hi @jquense. This pr is important. Please merge. Thanks!

